PR #39, opened 2014-11-26
Expose Headers, Request and Response to window

othree:
These three constructor should expose to global (window and worker).
Especially Headers.
Because headers in fetch option accepts native object and Header object.


josh:
FF has it exposed now too. Merging so we can start writing tests against the constructors.

PR #54, opened 2015-01-05
Mark FormData support as optional

josh:
Hypothetically, if a browser doesn't not implement `FormData` but did implement `fetch()`, how should `Response.prototype.formData` behave? Should `Response.prototype.formData` be `undefined` or should calling `Response.prototype.formData` throw some sort of error?

Chrome Canary currently does not implement `Response.prototype.formData` for an example. Also, old polyfilled IEs won't have FormData.

/cc @dgraham @mislav @annevk


dgraham:
I wrote the FormData support for completeness with the spec. We actually have no use cases for it on github.com.

I guess I don't understand the hypothetical situation described. All browsers with `fetch` also have `FormData`. Is this a question specifically for IE 9?

josh:
The same applies to `blob`. PhantomJS doesn't implement `Blob()` so it skips defining `Response.prototype.blob`.

dgraham:
Ah, PhantomJS is the new IE 9.

josh:
After this PR, all *supported* tests pass in Chrome Canary.

annevk:
If there's no support I would expect the property to not be there so you can feature test it.

PR #55, opened 2015-01-05
Test Atomic HTTP redirect handling

### LINKED ISSUE #42: Add test for HTTP redirect behavior

josh:
https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

Should behave just like XHR's redirection.

@annevk is `Response.prototype.url` supposed to be the redirected location? That was one of my feature wish list items that XHR can't do.

Closes #42.

/cc @dgraham @annevk

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,23 @@ fetch('/users')
   })
 ```
 
+### Response URL caveat
+
+The `Response` object has a URL attribute for the final responded resource.
+Usually this is the same as the `Request` url, but in the case of a redirect,
+its all transparent. Newer versions of XHR include a `responseURL` attribute
+that returns this value. But not every browser supports this. The compromise
+requires setting a special server side header to tell the browser what URL it
+just requested (yeah, I know browsers).
+
+``` ruby
+response.headers['X-Request-URL'] = request.url
+```
+
+If you want `response.url` to be reliable, you'll want to set this header. The
+day that you ditch this polyfill and use native fetch only, you can remove the
+header hack.
+
 ## Browser Support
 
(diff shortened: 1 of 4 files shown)


josh:
Gah, PhantomJS doesn't implement 308.

dgraham:
> is Response.prototype.url supposed to be the redirected location?

[Section 4.2](https://fetch.spec.whatwg.org/#http-fetch) steps 4 and 10 say yes.

annevk:
@josh XMLHttpRequest has `responseURL` these days. (And yes, it is the URL of the of the actual response.)

josh:
> XMLHttpRequest has responseURL these days. 

Nice! but looks like Safari still doesn't. I doubt IE does either. I wonder what we should set `response.url` to in the case that `responseURL` is `undefined`.

In a pre-responseURL world, github.com sends back a `X-Request-URL` header on every XHR. Then we can `getResponseHeader` that value. wdyt about this polyfill suggesting a similar fallback?

1. Try `responseURL`
2. Try `getResponseHeader('X-Request-URL')
3. Fallback to the originally requested url. Will be wrong if a redirect happened, but mostly right otherwise.

josh:
All tests pass on Chrome Canary (native fetch), Chrome and Firefox. Currently failing on Safari and IE because of missing `responseURL`.

dgraham:
I'm hesitant to move some of our github.com header logic into the polyfill. The native fetch implementations won't have that header, so polyfill users could become dependent on non-standard behavior that disappears when browsers enable `window.fetch` natively.

josh:
> The native fetch implementations won't have that header

Native implementations already do the right thing. The header does nothing in that case.

PR #56, opened 2015-01-06
Support credentials

### LINKED ISSUE #43: Add test for HTTP cookies

josh:
Incomplete.

I'm not sure how to support `credentials: 'omit'` with XHR.

Closes #43.

/cc @dgraham @annevk @mislav

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -197,6 +197,16 @@ promiseTest('supports HTTP DELETE', 2, function() {
   })
 })
 
+test('request credentials defaults to omit', function() {
+  var request = new Request('')
+  equal(request.credentials, 'omit')
+})
+
+test('request credentials uses inits member', function() {
+  var request = new Request('', {credentials: 'same-origin'})
+  equal(request.credentials, 'same-origin')
+})
+
 promiseTest('does not send cookies with implicit omit credentials', 1, function() {
   return fetch('/cookie?name=foo&value=bar').then(function(response) {
     return fetch('/cookie?name=foo');
--- a/test/test.js
+++ b/test/test.js
@@ -197,7 +197,7 @@ promiseTest('supports HTTP DELETE', 2, function() {
   })
 })
 
-promiseTest('doesnt send cookies with implicit omit credentials', 1, function() {
+promiseTest('does not send cookies with implicit omit credentials', 1, function() {
   return fetch('/cookie?name=foo&value=bar').then(function(response) {
     return fetch('/cookie?name=foo');
   }).then(function(response) {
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -196,3 +196,43 @@ promiseTest('supports HTTP DELETE', 2, function() {
     equal(request.data, '')
   })
 })
+
+promiseTest('doesnt send cookies with implicit omit credentials', 1, function() {
```
dgraham: "does not"

dgraham:
We need to default to "omit" in the [Request constructor](https://github.com/github/fetch/blob/9b1ddc9ef47011c2c9b0c9bc89e7ec09cf5ca962/fetch.js#L136).

I wish the default was `same-origin` rather than `omit`. All of our fetch usage requires cookie headers.

josh:
Any ideas how to disable cookies for `'omit'`? It seems like `withCredentials` is always ignored for same origin requests.

annevk:
That feature does not exist in XMLHttpRequest. XMLHttpRequest only has same-origin and CORS.

I could be convinced to have the standard default to same-origin, but it seems rather ugly to have a different default from cross-origin. (Mistake with how we designed XMLHttpRequest...)

josh:
@annevk it seems like the majority of fetch contexts (image, script, style, subresources) mostly use `"include"` even for cross origin resources with the exception of XHR's CORS semantics. I kinda agree that world would be a better place if stuff like `script` was actually `omit` by default. But for most direct `fetch()` api calls, I think I'll be using `same-origin` most the time.

annevk:
Well they use include but also set mode to no-cors by default. If you use the crossorigin attribute I think the intent is omit and mode cors, though implementations might do same-origin and mode cors...

dgraham:
> I could be convinced to have the standard default to same-origin

@annevk From my perspective, a default of `same-origin` is most useful. A default of `omit` means that `fetch` cannot be used easily by sites with user sessions maintained by cookie headers.

A simple `fetch('/some/url')` call becomes `fetch('/some/url', {credentials: 'same-origin'})` at *each* call site. I'd prefer to specify `omit` in the few cases it's needed instead.

For now, we've [added a default of `same-origin`](https://gist.github.com/dgraham/92e4c45da3707a3fe789#file-fetch-js-L37-L45) to the `$.fetch` wrapper we use on github.com.

If you think it's appropriate, maybe the default in the specification could be changed to `same-origin`?

annevk:
I recommend filing a bug against the specification. I don't really like `same-origin` myself as what it does depends on the origin of the URL you passed in, which is super weird.

PR #58, opened 2015-01-10
Make fetch.js pass the Closure Compiler

sayrer:
These annotations are needed to make it compile without error--more could be added to increase correctness.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -5,6 +5,10 @@
     return
   }
 
+  /**
+   * @param {Headers|Object=} headers
+   * @constructor
+   */
   function Headers(headers) {
     this.map = {}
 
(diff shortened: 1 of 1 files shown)


dgraham:
Thanks, but closure compiler isn't part of the development workflow for this library. It also seems strange that Java comment annotations are required in JavaScript code for this tool.

sayrer:
Well, lack of these comments make it hard to use in projects with the Closure Compiler's strict type checking--not saying this project has to use it.

The annotations are JSDoc fwiw, and a lot of libraries provide them. Here's an example from React:

https://github.com/facebook/react/blob/78ec2501cf9d57fa38d3166ceae0e9b7811071c2/src/utils/OrderedMap.js#L62

necolas:
This is basic JSDoc (not Java comments), as supported by several Node.js linting tools. Would be useful to many people to include these annotations, especially companies using Closure Compiler.

PR #60, opened 2015-01-11
Support fetching binary data using xhr.responseType = 'blob'

lrowe:
Perhaps it should request with `xhr.responseType = 'arraybuffer'` and convert to a string manually in `text()`.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -73,29 +73,56 @@
     this._body = null
     this.bodyUsed = false
 
-    this.arrayBuffer = function() {
-      var rejected = consumed(this)
-      if (rejected) {
-        return rejected
+    if (has_responseType) {
+      this.arrayBuffer = function() {
+        var rejected = consumed(this)
+        if (rejected) {
+          return rejected
+        }
+
+        var body = this._body
+        return new Promise(function(resolve, reject) {
+          var fileReader = new FileReader()
+          fileReader.readAsArrayBuffer(body)
+          fileReader.onloadend = function(e) {
+            if (e.target.error) {
+              reject(e.target.error)
+            } else {
+              resolve(e.target.result)
+            }
+          }
+        })
       }
 
-      var body = this._body
-      return new Promise(function(resolve, reject) {
-        var fileReader = new FileReader()
-        fileReader.readAsArrayBuffer(body)
-        fileReader.onloadend = function(e) {
-          if (e.target.error) {
-            reject(e.target.error)
-          } else {
-            resolve(e.target.result)
-          }
-        }
-      })
-    }
+      this.blob = function() {
+        var rejected = consumed(this)
+        return rejected ? rejected : Promise.resolve(this._body)
+      }
 
-    this.blob = function() {
-      var rejected = consumed(this)
-      return rejected ? rejected : Promise.resolve(this._body)
+      this.text = function() {
+        var rejected = consumed(this)
+        if (rejected) {
+          return rejected
+        }
+
+        var body = this._body
+        return new Promise(function(resolve, reject) {
+          var fileReader = new FileReader()
+          fileReader.readAsText(body)
+          fileReader.onloadend = function(e) {
+            if (e.target.error) {
+              reject(e.target.error)
+            } else {
+              resolve(e.target.result)
+            }
+          }
+        })
+      }
+    } else {
+      this.text = function() {
+        var rejected = consumed(this)
+        return rejected ? rejected : Promise.resolve(this._body)
+      }
     }
 
     if (self.FormData) {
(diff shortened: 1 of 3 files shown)


lrowe:
This is failing due to ariya/phantomjs#11013 which is only fixed for phantomjs 2.0.

fetch.js
```diff
@@ -68,58 +68,85 @@
...
+      var match = charset_re.exec(this.headers.get('Content-Type') || '')
+      var charset = match ? match[1] : null
       var body = this._body
       return new Promise(function(resolve, reject) {
-        try {
-          resolve(JSON.parse(body))
-        } catch (ex) {
-          reject(ex)
+        var fileReader = new FileReader()
+        fileReader.readAsText(body, charset)
```
josh: I think only utf-8 decoding `text()` will be supported.

> Resolve p with the result of running utf-8 decode on bytes.

https://fetch.spec.whatwg.org/#body-mixin

josh:
I'm a little worried about requiring XHR2 features like `responseType`. I don't want to jettison IE9 completely. I think the `.blob` mixin is the only one thats going to support binary data, text and json should still work fine given the utf-8 only constrain.

lrowe:
Perhaps the implementation should switch based on `'responseType' in (new XMLHttpRequest()` leaving IE8/9 using responseText?
The workarounds for accessing binary data in ie8 look pretty nasty: http://miskun.com/javascript/binary-file-loading-in-javascript/. It might be possible to make something work combined with a typedarray polyfill like: https://bitbucket.org/lindenlab/llsd/src/default/js/typedarray.js but I'm not sure its worth the hassle.

lrowe:
I've removed the charset handling per spec and use feature detection on 'responseType' to switch between implementations. All tests except for the already failing 308 redirect pass on IE9.

josh:
> Perhaps the implementation should switch based on 'responseType' in (new XMLHttpRequest() leaving IE8/9 using responseText?

Yeah, that sounds pretty good to me. Don't worry about IE8 though.

fetch.js
```diff
@@ -194,6 +226,9 @@
       }
 
       xhr.open(self.method, self.url)
+      if (has_responseType) {
```
josh: I'd probably just do the feature detect inline `'responseType' in xhr`

test/test.js
```diff
@@ -63,6 +63,36 @@ suite('Response', function() {
 
 // https://fetch.spec.whatwg.org/#body-mixin
 suite('Body mixin', function() {
+  ;(Response.prototype.blob ? suite : suite.skip)('arrayBuffer', function() {
```
josh: Can we feature detect `Response.prototype.arrayBuffer` here?

test/server.js
```diff
@@ -28,6 +28,10 @@ var routes = {
     });
     res.end('hi');
   },
+  '/nonascii': function(res) {
```
josh: Maybe just return some binary bytes from this test endpoint? Maybe even something thats invalid utf-8 just for good coverage.

fetch.js
```diff
@@ -73,60 +73,92 @@
...
-        } catch (ex) {
-          reject(ex)
-        }
+      return this.text().then(function(text) {
+        return new Promise(function(resolve, reject) {
+          try {
+            resolve(JSON.parse(text))
+          } catch (ex) {
+            reject(ex)
+          }
```
josh: You don't need to make a new Promise here.

Just do `this.text().then(function(text) { return JSON.parse(text) })`

fetch.js
```diff
@@ -73,60 +73,92 @@
...
-        return rejected ? rejected : Promise.resolve(decode(this._body))
+        return this.text().then(function(text) {
+          return new Promise(function(resolve, reject) {
+            try {
+              resolve(decode(text))
+            } catch (ex) {
+              reject(ex)
+            }
+          })
+        })
```
josh: ditto, `this.text().then(function(text) { return decode(text) })`

fetch.js
```diff
@@ -73,60 +73,92 @@
     this._body = null
     this.bodyUsed = false
 
-    this.arrayBuffer = function() {
-      throw new Error('Not implemented yet')
-    }
+    if (has_responseType) {
```
josh: For IE9, checking for `FileReader` support should cover this definitions.

josh:
> I'm not sure what to do about phantomjs.

Does phantom return a false positive for the `responseType` feature check? Trying to avoid the phantom UA checks, but I think some of the other checks should force it into the IE9 code path.

josh:
@lrowe thanks for getting this going. Definitely a good direction here for handling binary data.

fetch.js
```diff
@@ -73,60 +73,92 @@
...
+      this.arrayBuffer = function() {
```
josh: Maybe array buffer mixin is `ArrayBuffer && FileReader`

fetch.js
```diff
@@ -73,60 +73,92 @@
...
+            } else {
+              resolve(e.target.result)
+            }
+          }
+        })
       }
-    })();
 
-    if (blobSupport) {
       this.blob = function() {
```
josh: We'll still want a blob feature check.

lrowe:
> Does phantom return a false positive for the responseType feature check? Trying to avoid the phantom UA checks, but I think some of the other checks should force it into the IE9 code path.

Looks like its Blob type is broken but is otherwise based on a fairly recent WebKit.

fetch.js
lrowe: Happy to change this, but note the module scope feature detection would still be required for switching method implementations in `Body`.

josh:
> Looks like its Blob type is broken but is otherwise based on a fairly recent WebKit.

Yeah, thats why I had the blob constructor feature check. `var blobSupport = (function() {`

PR #62, opened 2015-01-11
Make Headers case insensitive though lowercasing.

lrowe:
This implementation is not case preserving when iterating over a Headers object.

See: https://fetch.spec.whatwg.org/#concept-header-name


josh:
Rad. Passes in Canary too.

PR #63, opened 2015-01-12
Suspect this api key was wrong

matthew-andrews:


josh:
yay! https://www.npmjs.com/package/whatwg-fetch

PR #65, opened 2015-01-13
Check for xhr.response only when supported.

kruppel:
ref: https://github.com/github/fetch/commit/74fa62be3dba24ea77f87d7aa5ab00ab90685b83

The above commit added a check for blob support to access the `response` property of the xhr. While that works for most cases, I'd argue the availability of XMLHTTPRequest2 and Blob are orthogonal and be treated as such.

For a practical use case, I haven't found a great solution for mocking XHR2 requests (generally use [Sinon.JS](https://github.com/cjohansen/Sinon.JS/issues/569)). Without the proposed change, any call to get a response body would return `undefined`. You can also imagine that any consumer who polyfills `Blob` in IE9 will run into the same type of issue.

Is there any harm in checking property existence in this instance?

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -199,7 +199,7 @@
           headers: headers(xhr),
           url: xhr.responseURL || xhr.getResponseHeader('X-Request-URL')
         }
-        resolve(new Response(blobSupport ? xhr.response : xhr.responseText, options))
+        resolve(new Response('response' in xhr ? xhr.response : xhr.responseText, options))
       }
 
       xhr.onerror = function() {


lrowe:
Looks good to me.

josh:
Seems fine.

> I'd argue the availability of XMLHTTPRequest2 and Blob are orthogonal and be treated as such.

Yeah, could we add a `'responseType' in xhr && blobSupport` check before setting `xhr.responseType = 'blob'`.

kruppel:
> Yeah, could we add a 'responseType' in xhr && blobSupport check before setting xhr.responseType = 'blob'.

XMLHttpRequest Level 1 should still have a `responseType` property. It looks like [Modernizr uses the following for feature detection](https://github.com/Modernizr/Modernizr/commit/ed2ecea1e5747788a2865e46f95d7e1c673eb7d6):

```javascript
var supportsXHR2 = 'withCredentials' in new XMLHttpRequest();
```

From the looks of [MDN's compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#Browser_Compatibility) though, I'm not sure that would cover all desired cases, i.e. some response types are unavailable until later browser versions.

In any case, more than happy to add whatever's decided best. Thanks for quick response.

dgraham:
> You can also imagine that any consumer who polyfills Blob in IE9

This usage scenario is not supported.

josh:
I still think this makes sense to logically check for.

lrowe:
It isn't really safe to pass the responseText string to the Response constructor when blobSupport is true. Somehow it needs to trigger the text only support. Maybe there should be a separate LegacyResponse object used when either blobSupport is false or using XHR1

Issue #61, opened 2015-01-11
`blob()` does not work with non-ascii data


lrowe:
Fetching non-ascii binary data corrupts the data due to responseText string decoding.

Failing test demonstrating problem in #60.

PR #66, opened 2015-01-14
Avoid conversion to blob when responseText used.

lrowe:
String decoding / encoding is likely to render the binary data corrupt so throw error instead.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -106,6 +106,9 @@
 
     if (blobSupport) {
       this.blob = function() {
+        if (!this._bodyBlob) {
+          throw new Error('XMLHttpRequest missing blob support')
+        }
         var rejected = consumed(this)
         return rejected ? rejected : Promise.resolve(this._bodyBlob)
       }
@@ -113,15 +116,13 @@
       this.arrayBuffer = function() {
         return this.blob().then(readBlobAsArrayBuffer)
       }
-
-      this.text = function() {
+    }
+    this.text = function() {
+      if (this._bodyBlob) {
         return this.blob().then(readBlobAsText)
       }
-    } else {
-      this.text = function() {
-        var rejected = consumed(this)
-        return rejected ? rejected : Promise.resolve(this._bodyText)
-      }
+      var rejected = consumed(this)
+      return rejected ? rejected : Promise.resolve(this._bodyText)
     }
 
     if ('FormData' in self) {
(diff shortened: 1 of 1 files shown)


fetch.js
```diff
@@ -106,6 +106,9 @@
 
     if (blobSupport) {
       this.blob = function() {
+        if (!this._bodyBlob) {
+          throw new Error('XMLHttpRequest missing blob support')
+        }
```
josh: I don't think this is correct. A `Response` can be constructed with a String and accessed from the blob mixing.

josh:
For response/body's `bodyInit`, its recommend that the value be read into a stream first.

https://fetch.spec.whatwg.org/#concept-bodyinit-extract

Eventually when whatwg streams are furthered spec'd they'll be decoded into the string instead of the blob.

josh:
Added some additional tests to master to show you what I mean 9d586486e50a79551b1d12178b3408d1fd57cb35

lrowe:
This is #61 again. IE 8/9 in combination with a Blob polyfill will corrupt binary data. It seems better to disallow usage of blob/arrayBuffer in that case.

To make the `Response` constructor compatible maybe the xhr.responseText could be wrapped in a special object to indicate that returning a blob is disallowed?

josh:
What blob polyfill are you using?

lrowe:
I'm not using a blob polyfill myself, though from #65 it looks like @kruppel is. I'd just like to rule out any possibility of data corruption.

josh:
If the polyfill Blob constructor doesn't work, I'm not sure what good it is.

lrowe:
I guess it could be useful for processing binary data in other ways, but it does not fix XMLHttpRequest for you. For that you would need the ugly VBScript work around.

lrowe:
Probably worth noting somewhere that with XMLHttpRequest 1 or without Blob support the charset will be auto-detected from the content-type header whereas spec is to always use utf8. So there's an argument for disabling `text()` on IE8/9 too, but that's probably going too far ;)

fetch.js
```diff
@@ -228,7 +229,9 @@
       options = {}
     }
 
-    if (blobSupport) {
+    if (bodyInit instanceof XMLHttpRequest) {
```
dgraham: This is non-standard behavior.

dgraham:
This branch is speculative—let's deal with errors as they arise. Blob usage is uncommon right now.

PR #67, opened 2015-01-14
Add metadata for repository, bugs and license

danez:
This would be nice as this information is displayed in the npm repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,19 @@
   "name": "whatwg-fetch",
   "version": "0.6.0",
   "main": "fetch.js",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/github/fetch.git"
+  },
+  "bugs": {
+    "url": "https://github.com/github/fetch/issues"
+  },
+  "licenses": [
+    {
+      "type": "MIT",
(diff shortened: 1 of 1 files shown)


domenic:
bugs is redundant with repository and repository can just be `"repository": "github/fetch"`

danez:
Oh I haven't checked that in detail, will do it later.

danez:
@domenic I couldn't find any reference on what you said, all examples state, you should us a full url to your repository. And bugs is a different url that is used by ```git bugs```

domenic:
It's correct, but it's way longer than it needs to be, and a maintenance burden.

Here's another official package or three you can copy markup from:

- https://github.com/domenic/webidl-class-generator/blob/master/package.json#L9
- https://github.com/domenic/sinon-chai/blob/master/package.json#L15
- https://github.com/tmpvar/jsdom/blob/master/package.json#L57

domenic:
And here's the docs: https://github.com/npm/normalize-package-data#what-normalization-currently-entails

danez:
Seems this shorter way of defining the guthub repository is not covered in the official docs about package.json.
Now its changed.

josh:
cools

PR #68, opened 2015-01-15
Timeout - candidate for spec

wheresrhys:
Proxies to the `xhr.timeout` property

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: node_js
 script: make test
-# deploy:
-#   provider: npm
-#   email: matt@mattandre.ws
-#   api_key:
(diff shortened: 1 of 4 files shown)


wheresrhys:
ok - I'll uncomment

matthew-andrews:
:+1:

dgraham:
We need the [Fetch spec](https://fetch.spec.whatwg.org) to describe standard timeout behavior before adding it to the polyfill. I agree it would be useful. You can open an issue at https://github.com/whatwg/fetch.

matthew-andrews:
oops, i think this pull request was meant for our internal fork of fetch.

PR #69, opened 2015-01-15
Fetch spec requires non-omit credentials mode to set cookies too.

nikhilm:
HTTP Network or cache fetch step 13 gates on the credentials flag.

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -98,7 +98,7 @@ var routes = {
   '/cookie': function(res, req) {
     var setCookie, cookie
     var params = querystring.parse(url.parse(req.url).query);
-    if (params.value && params.value) {
+    if (params.name && params.value) {
       setCookie = [params.name, params.value].join('=');
     }
     if (params.name) {
--- a/test/test.js
+++ b/test/test.js
@@ -451,14 +451,41 @@ suite('Atomic HTTP redirect handling', function() {
 suite('credentials mode', function() {
   var omitSupported = !self.fetch.polyfill
 
+  suite('clear cookies', function() {
+    test('clear cookie', function() {
+      return fetch('/cookie?name=foo&value=reset', {credentials: 'same-origin'});
+    })
+  })
+
   ;(omitSupported ? suite : suite.skip)('omit', function() {
     test('request credentials defaults to omit', function() {
       var request = new Request('')
       assert.equal(request.credentials, 'omit')
     })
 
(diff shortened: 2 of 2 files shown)


nikhilm:
Sorry for the merge commit :/

josh:
Yeah, this sounds right.

> If credentials flag is set and response's header list contains one or more headers named `Set-Cookie`, run these substeps:

Could @annevk confirm?

There would be a bug in Chrome Canary in this case cause it does pass here when it shouldn't. @nikhilm could you write an extra test that asserts that `Set-Cookie` is ignored in the credentials omit case.

test/server.js
```diff
@@ -98,7 +98,7 @@ var routes = {
   '/cookie': function(res, req) {
     var setCookie, cookie
     var params = querystring.parse(url.parse(req.url).query);
-    if (params.value && params.value) {
+    if (params.name && params.value) {
```
josh: ha, thanks

test/test.js
```diff
@@ -451,14 +451,41 @@ suite('Atomic HTTP redirect handling', function() {
...
+        return fetch('/cookie?name=foo', {credentials: 'same-origin'});
+      }).then(function(response) {
+        return response.text()
+      }).then(function(data) {
+        assert.equal(data, 'reset')
+      })
+    })
+
+    test('does not accept cookies with omit credentials', function() {
+      document.cookie = "";
```
josh: tests are probably mad at the double quotes on this line.

test/test.js
```diff
@@ -451,14 +451,41 @@ suite('Atomic HTTP redirect handling', function() {
 suite('credentials mode', function() {
   var omitSupported = !self.fetch.polyfill
 
+  suite('clear cookies', function() {
+    test('clear cookie', function() {
+      return fetch('/cookie?name=foo&value=reset', {credentials: 'same-origin'});
+    })
+  })
```
josh: Are the other tests depending on this running first to clear the cookies?

If so, we might want to do it as a suite setup function so it will always run first even if individual tests are ran standalone.

I think you basically do

``` js
setup(function() {
  return fetch(...)
})
```

josh:
Thanks @nikhilm!

annevk:
Confirmed, cookies should only be set if the request allowed credentials. Otherwise e.g. a CORS preflight could result in cookies being set.

josh:
Thanks @annevk!

PR #70, opened 2015-01-15
Improve Request/Response BodyInit consuming

josh:
Both Request and Response can accept a number of BodyInit types and are supposed to be consumable via a number of mixin readers.

@annevk seems like we're missing a way to consume `FormData` types. Theres no way to turn it into a Blob or ArrayBuffer. `new Response(formData).blob()` is going to become a clever hack if nothing official defined.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -109,9 +109,9 @@
         this._bodyInit = body
         if (typeof body === 'string') {
           this._bodyText = body
-        } else if (Blob.prototype.isPrototypeOf(body)) {
+        } else if ('Blob' in self && Blob.prototype.isPrototypeOf(body)) {
           this._bodyBlob = body
-        } else if (FormData.prototype.isPrototypeOf(body)) {
+        } else if ('FormData' in self && FormData.prototype.isPrototypeOf(body)) {
           this._bodyFormData = body
         } else if (!body) {
           this._bodyText = ''
@@ -158,6 +158,8 @@
         this._bodyInit = body
         if (typeof body === 'string') {
           this._bodyText = body
+        } else if ('FormData' in self && FormData.prototype.isPrototypeOf(body)) {
+          this._bodyFormData = body
         } else if (!body) {
           this._bodyText = ''
         } else {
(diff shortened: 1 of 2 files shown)


fetch.js
```diff
@@ -105,9 +105,34 @@
...
+        if (typeof body === 'string') {
+          this._bodyText = body
+        } else if (Blob.prototype.isPrototypeOf(body)) {
+          this._bodyBlob = body
+        } else if (FormData.prototype.isPrototypeOf(body)) {
+          this._bodyFormData = body
+        } else if (!body) {
+          this._bodyText = ''
+        } else {
+          throw new Error('unsupported BodyInit type')
```
josh: @annevk when handling `BodyInit` types, whats the correct way to handle unknown types? The spec just says "switch on object's type" with no default case.

Chrome Canary seems to just cast everything via `toString()`.

``` js
new Response(function() {}).text()
// "function () {}"
```

fetch.js
```diff
@@ -105,9 +105,34 @@
...
         var rejected = consumed(this)
-        return rejected ? rejected : Promise.resolve(this._bodyBlob)
+        if (rejected) {
+          return rejected
+        }
+
+        if (this._bodyBlob) {
+          return Promise.resolve(this._bodyBlob)
+        } else if (this._bodyFormData) {
+          throw new Error('could not read FormData body as blob')
```
josh: In the polyfill case, if you construct a Request/Reponse body with FormData then ask for the blob, we just crash :/

```
new Response(new FormData()).blob()
// throws an error
```

fetch.js
annevk: That is because IDL handles it. And indeed, if there's a primitive value present that is what IDL defaults to.

annevk:
@josh you can iterate over `FormData` these days. But yes, the serialization is not exposed. Part of the reason is that its specification is not quite clear. That needs to be sorted at some point.

josh:
``` js
(new FormData())[Symbol.iterator] // undefined
```

Doesn't look even Canary or FF Nightly implement it yet.

Anyway, thanks again @annevk! I'll just proceed with those cases being unsupported.

PR #72, opened 2015-01-16
HEAD/GET Request should throw if body is passed

nikhilm:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27846

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -334,7 +334,6 @@ suite('Methods', function() {
   test('supports HTTP GET', function() {
     return fetch('/request', {
       method: 'get',
-      body: 'name=Hubot'
     }).then(function(response) {
       return response.json()
     }).then(function(request) {
@@ -342,7 +341,17 @@ suite('Methods', function() {
       assert.equal(request.data, '')
     })
   })
-    test('supports HTTP POST', function() {
+
+  test('HTTP GET cannot have a body', function() {
+    try {
+      var r = new Request('', { body: 'hello' })
+      assert(false, "Should throw")
+    } catch(e) {
+      assert(true, "Should throw")
+    }
+  })
+
+  test('supports HTTP POST', function() {
     return fetch('/request', {
       method: 'post',
       body: 'name=Hubot'


josh:
I really hope this is something that will be supported when you consider `URLSearchParams` as an allowed `BodyInit` type. Serializing a form as `URLSearchParams` then issuing a GET fetch request is pretty useful.

/cc @annevk

dgraham:
Maybe we need some clarity around step 21 of the [Request constructor](https://fetch.spec.whatwg.org/#request-class) in the spec. It doesn't mention method constraints on body yet.

annevk:
@dgraham see OP. @josh why would you pass `URLSearchParams` as body rather than concatenated with the URL?

josh:
This is a pretty common approach to serializing a form into an XHR/fetch.

``` js
form.addEventListener('submit', => {
  fetch(form.action, { method: form.method, body: serializeForm(form) })
})
```

where `serializeForm` may return a `URLSearchParams`, `FormData`, encoded string, whatever. Otherwise you have to deal switching logic based on if the method is GET or non-GET, then decide how to append to the existing URL (it may already have an existing query) and url encode the data.

``` html
<form method="GET" action="/search?foo=bar">
  <input type="text" name="q" value="Fetch">
</form>
```

I wish it just followed the same logic we have for forms and deal with appending to the existing url for me. I realize it might be confusing because the key is called "body" and yeah, GET can't actually have an HTTP body.

I guess I don't fully understand the security issues as discussed on in the IRC log.

annevk:
`<form>` is much higher-level and knows about method semantics. `fetch()` is not supposed to know about method semantics. If you pass a body and method is `GET`, ideally we would actually make that the body, even though that is forbidden by the HTTP specification (we won't implement rules for each HTTP method in existence). The problem with doing that however is that existing servers might not be expecting that and we have to preserve those invariants (breaking them is a potential security problem for those servers).

Perhaps we should have a new feature for this use case. Perhaps `formData` that manipulates the URL if method is `GET` or `HEAD` serves as the body otherwise. Making `body` magical like that does not seem like a good idea.

annevk:
Consider also that apart from the `method` argument you also need to special case the `body` argument as e.g. a `FormData` object cannot be serialized as part of a URL.

annevk:
If we really wanted this we could enable this by throwing for "no-cors" and upgrading "cors" to "cors-with-forced-preflight". However, that would not mutate the URL, but would actually have it as part of the body. I think I'd rather just throw for now though.

josh:
Roger that.

PR #73, opened 2015-01-18
README: add Travis, Bower, npm badges

cvrebert:


josh:
Thanks, but I think we're fine without badges.

PR #75, opened 2015-01-21
Include PATCH method

gdi2290:
no PATCH method love?

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -187,7 +187,7 @@
   }
 
   // HTTP methods whose capitalization should be normalized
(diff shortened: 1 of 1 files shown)


josh:
`PATCH` normalization isn't in the spec.

https://fetch.spec.whatwg.org/#methods

dgraham:
https://github.com/whatwg/fetch/issues/17

mislav:
See dcc3e82a8d5b4289c9bfa465b12b657df7c63cb5

PR #76, opened 2015-01-21
Add repository url to package.json

hawkrives:
npm has a habit of complaining when things aren't to its liking, like a package.json that lacks a `repository` field.

This uses npm's special-for-github `username/repo` syntax to be concise.


hawkrives:
... er, looks like this is pretty much covered in #67. Welp.

Sorry about the noise.

PR #77, opened 2015-01-23
If cors, with credentials

matthew-andrews:
Not really sure if I'm doing this right but this felt like it was missing.

PR #78, opened 2015-01-24
Centralise the checks for blob and form data support

matthew-andrews:
This helps with isomorphic-fetch (currently it's relying on a fork that has these extra few lines:- https://github.com/matthew-andrews/fetch/blob/master/fetch.js#L3-5) and (I think) makes fetch's implementation simpler.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -100,6 +100,7 @@
       return false
     }
   })();
+  var formDataSupport = 'FormData' in self;
 
   function Body() {
     this.bodyUsed = false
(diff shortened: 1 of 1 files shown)


fetch.js
```diff
@@ -109,9 +110,9 @@
         this._bodyInit = body
         if (typeof body === 'string') {
           this._bodyText = body
-        } else if ('Blob' in self && Blob.prototype.isPrototypeOf(body)) {
+        } else if (blobSupport && Blob.prototype.isPrototypeOf(body)) {
```
matthew-andrews: wasn't sure about this because `blobSupport` actually checks a lot more than just `'Blob' in self`

josh:
If it would help you more, I wouldn't mind centralizing all the checks to the top with a single object like

``` js
var support = {
  "Blob": howeverTestBlob,
  "FileReader": somethingTestFileReader
}
```

We could even export it as `fetch._support` if you wanted to runtime hack or inspect it.

matthew-andrews:
Umm, I don't mind.  I would just like it to do its checks upfront so that I can [restore `self`](https://github.com/matthew-andrews/isomorphic-fetch/blob/master/server.js#L17) after `require`ing fetch.

I can definitely tidy them all up into a single object though.

josh:
:heart:

matthew-andrews:
Would you mind if I made a (minor) release for this & https://github.com/github/fetch/pull/77?

josh:
no problem, done 115cd972a7d5dbc63bd2335a05976070e0ebc48c

PR #79, opened 2015-01-24
Add in the limited support IE9 offers for cross-origin requests

matthew-andrews:


josh:
Oh, fun :trollface: 

Thanks for getting this going. I'll probably need to take some time to test this all locally myself. A little busy next week.

I think we'll probably want to get CORS tests going at this point.

matthew-andrews:
This one is going to be especially fun to write tests for — SauceLabs?

Also I don't know JavaScript as well as you to know when not to use `;`s :disappointed:.

No rush for this and yes CORS testing…

domenic:
Don't do it. XDomainRequest is so far from XHR semantics you're going to be chasing bugs for years...

josh:
> This one is going to be especially fun to write tests for — SauceLabs?

Yeah, I really need to get that working too. CI is just using PhantomJS right now.

I'll probably steal the SauceLabs scripts I wrote for another project.

https://github.com/josh/async-form-element/blob/master/.travis.yml#L8-L13
https://github.com/josh/async-form-element/blob/master/test/saucelabs.js

> Also I don't know JavaScript as well as you to know when not to use ;s 

I'm actually in the pro semicolon camp :smirk:, but I didn't start this project.

> Don't do it. XDomainRequest is so far from XHR semantics you're going to be chasing bugs for years...

Haha. I personally don't need so I don't care too much about IE9 CORS support.

matthew-andrews:
> Haha. I personally don't need so I don't care too much about IE9 CORS support.

Sadly, we do :disappointed:.

josh:
> Sadly, we do :disappointed:.

Sorry, I didn't mean we couldn't investigate it. Especially if you'll be helping on the implementation.

dgraham:
The ideal place for this investigation to happen is a fork that matures on a production site. The maintenance burden of supporting non-standard APIs is too great for this mainline polyfill. Think of this repo as jQuery 2.x, supporting recent browsers, and the fork as jQuery 1.x, providing compatibility to old browsers.

We need to avoid exposing github.com traffic to this, where we have no way of properly testing it, because we don't make cross-origin requests from IE 9. Unless @josh has time to test and maintain this in the future, I think we should close this out.

josh:
> The ideal place for this investigation to happen is a fork that matures on a production site. 

@matthew-andrews is working on this for production @ftlabs apps.

matthew-andrews:
Yup.  This'll go on the new responsive version of FT.com that we're working on right now.  You'll be glad to hear that IE9 is as far back as we have to go (which the fetch polyfill does [claim support of](https://github.com/github/fetch#browser-support) so I would argue that it's not *quite* the same as jQuery 1 vs. 2) but if you'd prefer not to merge it, that's cool—we can use my fork.

I might make a PR to update this repo's `README.md` to warn about partial IE9 support, would that be OK?

josh:
I think if we can get a CORS suite done first (which we'll want no matter what), then see how complicated it would be to make it work in IE 9 and decide if its worth it. Maybe its super gnarly 100 lines or maybe its not too bad. I'm pretty sure @matthew-andrews is in to help maintain it if it makes it in core.

Does that sound like a good plan?

dgraham:
Closing per https://github.com/github/fetch/issues/176#issuecomment-122460601:

> . . . we ended up giving up on trying to make XDomainRequest work (it just wasn't good enough for us for any but the most basic cases)

PR #82, opened 2015-01-27
Add a convenience `ok` getter on `Response`

matthew-andrews:
[ https://github.com/whatwg/fetch/commit/e62f57b6c923f16e877ab56913a2a65388a7535a ]

Also wondering if also need to polyfill `ok` on the next version of Chrome where `fetch` might now be a partial implementation (sorry, google. my bad…)

---

Mozilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1126483

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -306,7 +306,7 @@
     this.type = 'default'
     this.url = null
     this.status = options.status
-    this.ok = this.status >= 200 && this.status < 300;
+    this.ok = this.status >= 200 && this.status < 300
     this.statusText = options.statusText
     this.headers = options.headers
     this.url = options.url || ''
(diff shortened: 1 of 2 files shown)


fetch.js
```diff
@@ -306,6 +306,7 @@
     this.type = 'default'
     this.url = null
     this.status = options.status
+    this.ok = this.status >= 200 && this.status < 300;
```
dgraham: Extra semicolon here.

dgraham:
We could wait to see if this appears in Chrome and Firefox nightlies before merging.

matthew-andrews:
Yep sure, that's fine by me.  Would you mind me keeping the PR open in the meantime?

dgraham:
Yep, let's keep this open!

josh:
:heart:

dgraham:
The `ok` attribute is now in Firefox and Chrome nightly builds.

Issue #80, opened 2015-01-27
npm-ignore files


necolas:
You can use `files` in `package.json` to specify the limited subset of project files that should be published to npm.

jonathanong:
looks like this can be closed. that PR was not renamed properly :D

PR #84, opened 2015-01-28
adds .npmignore

frankcash:
(#80)

Create .npmignore
#### Directories
`test/`
`examples/`

#### Files
`.npmignore`
`.gitignore`
`.jshintrc`
`.travis.yml`


domenic:
Better to use `files` property in `package.json`.

PR #85, opened 2015-01-29
Follow spec on Headers to throw TypeError, add tests for Headers

mattking17:


dgraham:
Chrome and Firefox nightlies currently accept non-String keys and values:

```javascript
h = new Headers()
h.append(12, 42)
// => undefined
h.has(12)
// => true
```

@annevk Is this correct behavior, or should a `TypeError` be thrown for non-String headers?

josh:
Nice!

It looks like Chrome native `fetch()` doesn't throw on bad types, but just `toString()`s them.

screen shot 2015-01-29 at 11 59 07 am

Think you can help report a bug on the Chrome issue tracker for this?

domenic:
The spec says to convert to string, not to throw.

domenic:
https://heycam.github.io/webidl/#es-ByteString

josh:
Ah interesting, these body tests maybe wrong too

https://github.com/github/fetch/blob/c3fc0d43a058944988eebb61b67c8a6549efcc9b/test/test.js#L385-L401

domenic:
Yes, they definitely are.

mattking17:
According to this it should throw TypeError https://fetch.spec.whatwg.org/#headers-class

domenic:
Which step?

mattking17:
Example:

To append a name/value (name/value) pair to a Headers object (headers), run these steps:
If name is not a name or value is not a value, throw a TypeError.

It's listed under each method definition.

domenic:
Right, so you should be testing whether they match https://tools.ietf.org/html/rfc2616#section-4.2, not doing a type-test. They're already converted to strings since the method signature contains ByteString/DOMString/USVString (depending on the method).

mattking17:
It's not clear to me what it should match based on the RFC, any ideas? Or is this check not really necessary?

annevk:
@dgraham that might not necessarily be a bug if both methods stringify. If the value of "12" is not a string you found a bug.

domenic:
@mattking17 you quoted

To append a name/value (name/value) pair to a Headers object (headers), run these steps:
If name is not a [name](https://fetch.spec.whatwg.org/#concept-header-name) or value is not a value, throw a TypeError.

The "name" link says

>  A name is a case-insensitive byte sequence that matches the [field-name](http://tools.ietf.org/html/rfc2616/#section-4.2) token production.

The "field-name" link links to the RFC which defines

```
field-name     = token
```

mattking17:
Thanks @domenic, that makes sense now. Updated the PR with field name checks to match the RFC, and removed string type checking and put in `toString` calls instead.

mattking17:
Let me know if there's anything else that this PR needs, more than happy to make necessary changes.

josh:
@mattking17 looks good. I'll get to testing and merging today. Thanks so much! :heart:

PR #86, opened 2015-02-03
filereader fix and form data change

nikhilm:


test/test.js
```diff
@@ -1,27 +1,15 @@
...
   return new Promise(function(resolve, reject) {
-    var reader = new FileReader()
-    reader.onload = function() {
-      resolve(reader.result)
-    }
-    reader.onerror = function() {
-      reject(reader.error)
-    }
-    reader.readAsText(blob)
+    var reader = new FileReaderSync()
```
josh: I don't think anyone has implemented `FileReaderSync` yet.

nikhilm: hmm.. gecko has had it for a while now. oh right, i completely forgot this code is shared by window :/

test/test.js
```diff
@@ -1,27 +1,50 @@
...
-    var reader = new FileReader()
-    reader.onload = function() {
-      resolve(reader.result)
-    }
-    reader.onerror = function() {
-      reject(reader.error)
+    // FileReader is not available on Workers in Firefox, but FileReaderSync
+    // is.
+    var reader
+    if (typeof FileReaderSync !== "undefined") {
```
stof: shouldn't you check for FileReader first instead, to use the async version when both are available ?

nikhilm:
I'm not sure why the travis build failed. I won't be able to fix this until next week, so feel free to steal.

test/test.js
```diff
@@ -1,27 +1,50 @@
...
   return new Promise(function(resolve, reject) {
-    var reader = new FileReader()
-    reader.onload = function() {
-      var view = new Uint8Array(reader.result)
-      resolve(Array.prototype.slice.call(view))
-    }
-    reader.onerror = function() {
-      reject(reader.error)
+    var reader
+    if (typeof FileReaderSync !== "undefined") {
```
stof: this should use single quotes (in both places) to fix the JSHInt warnings which are making the build fail

test/test.js
```diff
@@ -1,27 +1,50 @@
...
+      reader = new FileReaderSync()
+      try {
+        var view = new Uint8Array(reader.readAsArrayBuffer(blob))
+        resolve(Array.prototype.slice.call(view))
+      } catch(e) {
+        reject(e);
+      }
+    } else {
+      var reader = new FileReader()
```
stof: ``var`` should be removed here as it is already declared previously (see the Travis JSHint error)

test/test.js
```diff
@@ -1,27 +1,50 @@
 function readBlobAsText(blob) {
   return new Promise(function(resolve, reject) {
```
josh: We won't need any promise wrapping in the sync case, you can just define the function as

``` js
function readBlobAsText(blob) {
  return new FileReaderSync().readAsText(blob);
}
```

stof: This would make the return value of the ``readBlobAsText`` function inconsistent. Other parts of the library are expecting to get a promise from it

stof: ah no. It i indeed used only as a promise callback, so it is fine.

josh:
Thanks @nikhilm. I was able to cherry pick these changes into master with the linting tests fixed.

PR #87, opened 2015-02-04
fix server.js to server test.html to run in a browser

mattking17:
I wanted to run the tests in a browser, so I fixed it to serve test.html properly.

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -146,4 +146,4 @@ http.createServer(function(req, res) {
   }
 }).listen(port);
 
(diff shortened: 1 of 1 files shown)


josh:
The purpose of https://github.com/github/fetch/blob/master/test/server.js#L132 is to serve all these files statically.

mattking17:
Hi @josh, I just figured out that I can go to /test/test.html and it works as expected. Thanks!

Issue #91, opened 2015-02-28
Add support for file protocol (File URI Scheme)


ryaneof:
Hi there,
I was working on an Android project and using webview on some pages, and I was trying to use `fetch` to fetch some cached data on the device, which is able to access via `file:///` protocol. It seems `fetch` would throw a `TypeError` in this case for the moment.

dgraham:
Closing per https://github.com/github/fetch/pull/92#issuecomment-140665932.

PR #92, opened 2015-02-28
Add support for file protocol (File URI Schema)

ryaneof:
This commit would work while fetching a `file:///` protocol resource, mentioned in issue #91

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -279,8 +279,7 @@
       }
 
       xhr.onload = function() {
-        var protocol = /^([\w-]+:)\/\//.test(responseURL()) ? RegExp.$1 : self.location.protocol
-        var status = ((xhr.status === 1223) || (xhr.status === 0 && protocol === 'file:')) ? 204 : xhr.status
+        var status = ((xhr.status === 1223) || (xhr.status === 0 && /^([\w-]+:)\/\//.exec(responseURL())[1] === 'file:')) ? 204 : xhr.status
         if (status < 100 || status > 599) {
           reject(new TypeError('Network request failed'))
           return
--- a/fetch.js
+++ b/fetch.js
@@ -279,7 +279,8 @@
       }
 
       xhr.onload = function() {
-        var status = (xhr.status === 1223) ? 204 : xhr.status
+        var protocol = /^([\w-]+:)\/\//.test(responseURL()) ? RegExp.$1 : self.location.protocol
+        var status = ((xhr.status === 1223) || (xhr.status === 0 && protocol === 'file:')) ? 204 : xhr.status
         if (status < 100 || status > 599) {
           reject(new TypeError('Network request failed'))
           return


f15gdsy:
+1 for this pull.

charandas:
+1 for this pull.

mislav:
[Neither Chrome's nor Firefox's native implementation](https://github.com/github/fetch/issues/186#issuecomment-127479120) supports fetching local files. I think it was intentionally.

The fetch spec [doesn't explicitly forbid](https://fetch.spec.whatwg.org/#basic-fetch) the `file:` protocol, but it doesn't specify any details about it either, and is left "as an exercise to the reader".

I'm sorry, but I don't think fetch is meant for local files right now. You're better off using XHR.

ccapndave:
Just a note to point out that this is quite annoying when making hybrid apps (e.g. with Cordova) which regularly need access to local files.

screendriver:
+1 for Cordova Apps

mislav:
```js
function fetchLocal(url) {
    var xhr = new XMLHttpRequest
    xhr.onload = function() {
      resolve(new Response(xhr.responseText, {status: xhr.status})
    }
    xhr.onerror = function() {
      reject(new TypeError('Local request failed'))
    }
    xhr.open('GET', url)
    xhr.send(null)
  })
}
```

thw0rted:
I think this was mentioned elsewhere, but as of today Firefox supports fetch for `file://` URLs, when running in a `file://` origin.  I believe the restriction is that the requested file must be in or under the working directory where the top page is loaded from.

Chrome/Chromium still does not, but [I have submitted a bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=810400) because XHR does support it, with the appropriate command-line switch.

dineiar:
To summarize, [mislav's suggestion](https://github.com/github/fetch/pull/92#issuecomment-174730593) only works in Chrome with `--allow-file-access-from-files` flag (from [thw0rted's bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=810400)).

wbelhomsi:
inspired by @mislav
if using a framework that uses webpack or browserify put this in a script tag at the top of the index.html its not a foolproof solution but it does the job
```
(function() {
  var nativeFetch = window.fetch;
  if(nativeFetch) {
    var a = document.createElement('a');
    window.fetch = function(url) {
      a.href = url;
      var full_url = a.href;
      if(full_url.indexOf('file:') === 0) {
          var xhr = new XMLHttpRequest();
          xhr.onload = function() {
            resolve(new Response(xhr.responseText, {status: xhr.status}));
          };
          xhr.onerror = function(err) {
            reject(new TypeError('Local request failed'));
          };
          xhr.open('GET', url);
          xhr.send(null);
        });
      }
      else {
        return nativeFetch.call(this, Array.prototype.slice.apply(arguments));
      }
    };
  }
})();
```

PR #95, opened 2015-03-05
Use xhr.onreadystatechange for basic ie8 compatibility.

lrowe:
Use with es5-shim, es5-sham and es6-promise.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -278,7 +278,10 @@
         return;
       }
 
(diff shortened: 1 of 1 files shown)


dgraham:
See #96.

camsong:
I'm maintaining a fork `camsong/fetch-ie8` will support IE8 in the near future.

Issue #93, opened 2015-03-03
INVALID_STATE_ERR: DOM Exception 11 thrown by android browser.


joeandrews:

joeandrews:
I am getting the above exception thrown by the default android browser on Samsung Galaxy S4 w/ Android 4.3. I think it has something to do with async = true flag and a premature access to the response object.

joeandrews:
Its actually to do with cross domain request, same origin request are working fine, I will look into this further and if I find a fix submit a pull request.

PR #97, opened 2015-03-11
Set withCredentials after xhr.open for IE10 compatibility

adgad:
See http://stackoverflow.com/questions/19666809/cors-withcredentials-support-limited

Might also fix https://github.com/github/fetch/issues/93 but can't confirm

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -261,9 +261,6 @@
 
     return new Promise(function(resolve, reject) {
       var xhr = new XMLHttpRequest()
-      if (self.credentials === 'cors') {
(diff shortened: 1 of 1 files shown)


mislav:
Setting `withCredentials` after `open()` might fix some browsers but break others. We'd like to see some cross-browser tests. I would be a good idea to try to set up CORS automated testing as well, but this needn't be your responsbility in this PR.

iclanzan:
I don’t think this change breaks any browsers and without it requests with credentials fail in IE10 and IE11.

Issue #46, opened 2014-12-24
Make module available without polyfill


qubyte:
It would be nice to be able to require `fetch` without it being appended to `window` in a similar way to the `es6-promise` polyfill. If `fetch` can be assigned to a variable, then this means that when a broken or otherwise incompatible version of the fetch API become available in a browser, code that uses `fetch` won't break.

In other words:

```javascript
// This fetch can be used as is.
var fetch = require('fetch');

// Do this only if you want to polyfill fetch on the window object.
fetch.polyfill();
```

I realise however that this breaks the existing API.

josh:
https://github.com/github/fetch/issues/9#issuecomment-69297320

PR #98, opened 2015-03-14
Support RequireJS and Browserify (UMD)

moretti:
The main idea behind this PR is to fully support UMD.
`fetch` can still be included manually:
```
<script src="fetch.js"></script>
<script>
  fetch('/users.html').then(function () { /* ... */ });
</script>
```
but now it can also be required in an AMD or CommonJS environment:
```
var fetch = require('whatwg-fetch').fetch;
fetch('/users.html').then(function () { /* ... */ });
```
In the second case `fetch` doesn't patch the global environment automatically, but this can be done by calling `polyfill` if necessary:

```
require('whatwg-fetch').polyfill();
fetch('/users.html').then(function () { /* ... */ });
```

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,9 +1,15 @@
-(function() {
+(function (root, factory) {
   'use strict';
-
-  if (self.fetch) {
-    return
+  /* global module, define */
+  if (typeof define === 'function' && define.amd) {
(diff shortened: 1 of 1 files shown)


josh:
See #46

PR #99, opened 2015-03-16
Sauce Labs

josh:


mislav:
Much boilerplate :astonished:

test/test.js
```diff
@@ -35,7 +35,7 @@ test('resolves promise on 500 error', function() {
   })
 })
 
-test('rejects promise for network error', function() {
```
mislav: I C what you did there

josh:
@mislav any ideas for writing that one off ruby file in bash? I meant merge it into a single script.

mislav:
I usually do:

```bash
ruby -e '
  p STDIN.read.split("/")
'
```

You can even pipe in/out of that construct. Multiline ruby script will work fine, but if you quote it with single quotes outside then inside you can only do double quotes.

mislav:
> any ideas for writing that one off ruby file in bash? I plan merge it into a single script.

Heh, 6 months later: 07dc8ae4cc9a46ad4af35c99a8bdc0b83fbae28b

Issue #106, opened 2015-03-20
Request constructor should create Headers object when options.headers are passed


kirill-konshin:
```js
var r = new Response('{"foo":"bar"}', {headers: {'content-type': 'application/json'}});
r.headers.get // undefined
```

in `Response` constructor:

```js
this.headers = options.headers
```

should be:

```js
this.headers = new Headers(options.headers);
```

josh:
Fixed by #107. Thanks @kirill-konshin

PR #107, opened 2015-03-21
Headers constructor in Response constructor

kirill-konshin:
Fix for #106

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -327,7 +327,7 @@
     this.status = options.status
     this.ok = this.status >= 200 && this.status < 300
     this.statusText = options.statusText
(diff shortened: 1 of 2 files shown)


kirill-konshin:
I don't know why, but test run takes more than 50 minutes... Local tests are OK.

josh:
Does it pass in Chrome Canary with its native fetch implementation?

> I don't know why, but test run takes more than 50 minutes... Local tests are OK.

Don't worry about that for now. I'm still trying to get the sauce labs stuff working correctly.

kirill-konshin:
Yes, it does, I've checked in latest Chrome.

PR #110, opened 2015-03-26
explaned webpack usage

fourlastor:
Hi,

I noticed that both the NPM and Bower versions of fetch don't export fetch.

This is solvable by [shimming the module](https://github.com/webpack/docs/wiki/shimming-modules) as explained in webpack's documentation.

It might not be obvious to people tho (especially since an NPM package is provided), so I thought it would be sensible to add it to the readme, given the hype that webpack is having lately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,14 @@ If you want `response.url` to be reliable, you'll want to set this header. The
 day that you ditch this polyfill and use native fetch only, you can remove the
 header hack.
 
+### Webpack usage
+
+To use fetch with webpack, you will need `import-loader`, `export-loader`, `es6-promise` and require it this way:
+
+```javascript
+var fetch = require('imports?self=>{},es6p=es6-promise,Promise=>es6p.Promise!exports?self.fetch!whatwg-fetch');
+```
+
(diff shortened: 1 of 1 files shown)


nelix:
Will this stop the module from leaking into window?

fourlastor:
I think so, it creates (prepends code) a local variable into the file. So the window variable is actually shadowed by it.

nelix:
This will also allow me to leave the polyfilling of Promise to babel. yay.

mislav:
webpack

fourlastor:
I want that in the documentation too XD

README.md
```diff
@@ -154,6 +154,14 @@ If you want `response.url` to be reliable, you'll want to set this header. The
 day that you ditch this polyfill and use native fetch only, you can remove the
 header hack.
 
+### Webpack usage
+
+To use fetch with webpack, you will need `import-loader`, `export-loader`, `es6-promise` and require it this way:
```
ArmendGashi: Loader import-loader || export-loader' is not in the npm registry, name must be imports-loader exports-loader .

PR #112, opened 2015-04-02
coffeescript port

briice:
Hello,
I needed a Coffeescript version of this polyfill, so I forked yours. I don't know if it makes sense to merge it, but feel free if you think it is!

Brice


mislav:
Hi! Thanks for the suggestion, but we've decided from the start that we're going to author this polyfill in plain JS, and it doesn't make sense to us to maintain the library in 2 flavors.

Also, even if someone is writing all their site functionality in CoffeeScript, they are still able to load plain JS libraries like `fetch`. So we don't think it's a problem for anyone.

Take care!

Issue #109, opened 2015-03-25
credentials:"cors" not in spec


wereHamster:
```
dictionary RequestInit {
  // ...
  RequestMode mode;
  RequestCredentials credentials;
};

enum RequestMode { "same-origin", "no-cors", "cors" };
enum RequestCredentials { "omit", "same-origin", "include" };
```

You can use `{ mode: "cors" }` or `{ credentials: "include" }` but not `{ credentials: "cors" }`. I got confused why my code was not working because I was coding against the spec but then looked through the fetch.js source and saw that it is not compliant ;)

chromakode:
Just got bit by this one too. I think that [this check](https://github.com/github/fetch/blob/89b449384fe7e1858615d2ecf85c78802815d7bb/fetch.js#L264) should check `self.credentials === 'include'` instead of `'cors'`.

irvinebroque:
This issue is causing me pain too:

`credentials: 'cors',` works in browsers that have not yet implemented `fetch()`, but fails in Chrome Canary.
`credentials: 'include',` is spec-compliant, and works in Chrome Canary, but fails in all other browsers.

Are there blockers to getting either #113 or #97 merged in? Happy to help, but it seems like both PRs are stalled and it's unclear why.

dgraham:
This was introduced in #77. It's possible the spec changed since then. @matthew-andrews, any ideas on this?

alayor:
I'm having trouble with this site: http://www.greatapp.xyz/register
I receive XSRF-TOKEN but the Cookie it's not being set in Chrome.
I'm using `credentials: 'include'`

PR #113, opened 2015-04-02
Fix credentials include

chromakode:
Fixes https://github.com/github/fetch/issues/109.

It appears that cross-domain requests are not currently tested by the suite. However, the [test for this case](https://github.com/github/fetch/blob/89b449384fe7e1858615d2ecf85c78802815d7bb/test/test.js#L664) was not previously running since it was passing `credentials: 'include'`. With this change, the code actually runs, which was causing an exception in PhantomJS (see "Set xhr.withCredentials after xhr.open called.").

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -261,9 +261,6 @@
 
     return new Promise(function(resolve, reject) {
       var xhr = new XMLHttpRequest()
-      if (self.credentials === 'cors') {
-        xhr.withCredentials = true;
-      }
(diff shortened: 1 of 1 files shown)


chromakode:
Ah, looks like https://github.com/github/fetch/pull/97 contains the same change to `xhr.withCredentials`.

dgraham:
Closes #97.

PR #116, opened 2015-04-12
Support of passing the 'Request' object to the 'fetch' function as a first argument.

manifest:
The specification allows us to pass not only the URL as a first argument of the 'fetch' function but also the 'Request' object. This behavior has already been implemented in the actual version of Firefox.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -226,17 +226,18 @@
       input = {url: input}
     }
 
-    this.url = input.url;
+    this.url = input.url
     this.credentials = options.credentials || input.credentials || 'omit'
     this.headers = new Headers(options.headers || input.headers)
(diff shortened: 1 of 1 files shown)


fetch.js
```diff
@@ -337,7 +337,25 @@
...
+      options.credentials = options.credentials || input.credentials;
+      options.headers = options.headers || input.headers;
+      options.method = options.method || input.method;
+      options.mode = options.mode || input.mode;
+      options.body = options.body || input._bodyInit;
+    }
+    else {
+      url = input;
+    }
+
```
dgraham: The [`Request` constructor](https://fetch.spec.whatwg.org/#request-class) is responsible for this initialization.

dgraham:
Fixed in ab1ac93.

PR #117, opened 2015-04-13
Add `user` and `password` options and use for Basic Authentication

leungwensen:
reference: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#open()

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -225,6 +225,9 @@
     this.mode = options.mode || null
     this.referrer = null
 
+    this.user = options.user || null
+    this.password = options.password || null
+
     if ((this.method === 'GET' || this.method === 'HEAD') && options.body) {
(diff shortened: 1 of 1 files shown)


mislav:
Is this per-spec? I can't find evidence of `options.user` and `options.password` being defined in the spec. All I see is them being embedded in the URL.

This will need tests, too.

mislav:
Actually username/password in URLs is not supported either. #119

I don't know what's supported, then.

leungwensen:
It's not per-spec. Feel free to close this pull-request.

Actually I just want to keep some of my simple web services working after upgrade to window.fetch. Is there any plan to support `username:password` feature?

dgraham:
The `XMLHttpRequest` usage is an implementation detail that is not exposed in the `window.fetch` API.

PR #119, opened 2015-04-13
reject username:password in URL as per spec update

mattking17:
As per the spec update https://github.com/whatwg/fetch/issues/26

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -218,7 +218,7 @@
   function Request(url, options) {
     options = options || {}
 
-    if (url && url.match(/^\w+?:\/\/\w+?:\w+?@/)) {
+    if (url && url.match(/^([\w-]+:)?\/\/[^\/]+@/)) {
       throw new TypeError('Credentials not allowed in request URL')
     }
 
--- a/fetch.js
+++ b/fetch.js
@@ -217,6 +217,11 @@
 
   function Request(url, options) {
     options = options || {}
+
+    if (url && url.match(/^\w+?:\/\/\w+?:\w+?@/)) {
+      throw new TypeError('Credentials not allowed in request URL')
+    }
+
     this.url = url
 
     this.credentials = options.credentials || 'omit'
--- a/test/test.js
+++ b/test/test.js
@@ -139,6 +139,14 @@ suite('Request', function() {
     assert.equal(request.url, 'https://fetch.spec.whatwg.org/')
   })
 
+  test('rejects username:password in url', function() {
+    assert.throw(function() {
+      new Request('https://exampleuser:examplepass@fetch.spec.whatwg.org/', {
+        method: 'get'
+      })
+    }, TypeError)
+  })
+
   // https://fetch.spec.whatwg.org/#concept-bodyinit-extract
   suite('BodyInit extract', function() {
     ;(Request.prototype.blob ? suite : suite.skip)('type Blob', function() {


dgraham:
Has this made it into the [Fetch specification](https://fetch.spec.whatwg.org) yet?

mattking17:
Yes: https://fetch.spec.whatwg.org/#request-class
> If parsedURL includes credentials, throw a TypeError.

mislav:
Does the Fetch spec include any way of performing Basic Auth then, apart from manually constructing the `Authorization` header?

mattking17:
Also of note is inconsistent browser behavior with `username:password` in the URL (e.g. IE doesn't support it, at all).

It doesn't appear that you can set auth info in fetch other than using headers, though a simple function to produce the headers using `btoa`, or using a hardcoded string should do it.

mattking17:
Anything else I can do here? It does follow spec so not sure why we wouldn't include it.

mislav:
Patch looks :+1:, thanks for submitting.

This isn't a huge priority to fix so I'll let @dgraham merge at his convenience.

fetch.js
```diff
@@ -217,6 +217,11 @@
 
   function Request(url, options) {
     options = options || {}
+
+    if (url && url.match(/^\w+?:\/\/\w+?:\w+?@/)) {
+      throw new TypeError('Credentials not allowed in request URL')
+    }
+
```
dgraham: Using a regular expression to detect a little-used, edge case makes me worry about false positives. We should rely on the native `URL` class to handle parsing, and add a link to the [polyfill](https://github.com/components/url) in the README.md file for old browsers.

```javascript
parsed = new URL(url)
if (parsed.username || parsed.password) {
  throw new TypeError(...)
}
```

josh: Depending on `URL` seems legit to me.

mattking17:
It seems like more hoops to jump through to add another polyfill to support URL for like you said, a small edge case. Additionally, when I try to run tests in phantomjs, the `URL` object does not exist. Maybe there is a better way to detect username and password?

fetch.js
mislav: It's a bummer that we need to pull in a dependency just to be able to raise exceptions in rare cases. I think that a regex such as this would be just fine:

```js
/^([\w-]+:)?\/\/[^\/]+@/
```

mattking17:
Updated with suggested regex by @mislav

fetch.js
```diff
@@ -217,6 +217,11 @@
 
   function Request(url, options) {
     options = options || {}
+
+    if (url && url.match(/^([\w-]+:)?\/\/[^\/]+@/)) {
```
mislav: Looks alright now. Note that I'm not advocating use of regex to parse URLs in generally. For instance in cases when we had to detect cross-origin URLs in other libraries, using URL (or at least `HTMLAnchorElement#href` and related properties) was adamant to security (/cc @mastahyeti). However in this case I think it's relatively trivial to detect whether username or password have been specified, and raise an error appropriately. The worst thing that can happen is that this check fails and the XHR request made, but with user/password ignored because XHR never reads them from the URL string.

mastahyeti: > However in this case I think it's relatively trivial to detect whether username or password have been specified, and raise an error appropriately.

:smile: I disagree. This regex wouldn't match

- URLs with leading whitespace
- URLs with `+`, or `.` in the scheme.
- URLs with more than two `/` after the `:`
- Other valid, but pathological URLs.

Using `URL` would be much more robust.

> The worst thing that can happen is that this check fails and the XHR request made, but with user/password ignored because XHR never reads them from the URL string.

I'm not super familiar with the behavior of Fetch. If this isn't security sensitive, then a regex might be good enough.

mislav: > I'm not super familiar with the behavior of Fetch. If this isn't security sensitive, then a regex might be good enough.

Thanks for jumping in. Yes, this isn't security-sensitive. If you try to use a URL that contains username/password with XHR, the username/password gets ignored because XHR wants you to pass them as arguments to `xhr.open()`.

The Fetch spec contemplated supporting username/password in URL strings but dropped that idea because of the complexities around cross-origin redirects. So, this PR adds a helpful TypeError exception so you know what's goin on. If the URL string is super-weird and the regex fails, you simply won't get a helpful exception (and no auth credentials will be sent).

dgraham:
Closing this as stale for now. The extra error handling and spec compliance is appreciated. I'm just not comfortable matching URLs with a regular expression, and adding a dependency on the URL polyfill for IE makes using fetch in projects more difficult.

PR #120, opened 2015-04-13
remove un-needed promise allocations in example

calebboyd:


domenic:
Strong +1

README.md
```diff
@@ -117,10 +117,9 @@ resolved only on successful, 200 level, status codes.
 ```javascript
 function status(response) {
   if (response.status >= 200 && response.status < 300) {
```
matthew-andrews: I think you could also change this to `if (response.ok)`…

PR #124, opened 2015-04-17
Return early if self is undefined

tuespetre:
Return early if self is undefined, for environments that do not define 'self' (an example would be running a browserified script using ReactJS.NET.) This prevents a ReferenceError from occurring.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,7 +1,7 @@
 (function() {
   'use strict';
 
-  if (self.fetch) {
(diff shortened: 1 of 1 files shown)


josh:
The polyfill only designed to run in a browser environment, which is why we have the `self` requirement.

I'd probably suggest omitting it from your browserified bundle in that case.

mislav:
Please see #31

PR #130, opened 2015-04-29
Set "withCredentials" after ".open"

### LINKED ISSUE #129: IE10 throws on withCredentials
A trimmed down way to reproduce:
```javascript
try {
xhr.withCredentials = true;
}catch(e) {
    alert("There was an error:" + e);
}
```

More information: http://stackoverflow.com/questions/19666809/cors-withcredentials-support-limited

Moving:
```javascript
if (self.credentials === 'cors') {
    xhr.withCredentials = true;
}
```
until after the `xhr.open` fixes this for me.

Naddiseo:
Fixes #129

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -261,9 +261,6 @@
 
     return new Promise(function(resolve, reject) {
       var xhr = new XMLHttpRequest()
-      if (self.credentials === 'cors') {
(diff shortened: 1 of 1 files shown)


dgraham:
Fixed in #113.

PR #132, opened 2015-05-03
reject when request method is not a listed method

mattking17:
This covers step 20 on Request initialization:

> If init's method member is present, let method be it and run these substeps:
>    If method is not a method or method is a forbidden method, throw a TypeError.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -212,10 +212,7 @@
 
   function normalizeMethod(method) {
     var upcased = method.toUpperCase()
-    if (methods.indexOf(upcased) === -1) {
-      throw new TypeError('Method not allowed, must be one of ' + methods.join(', '))
-    }
-    return upcased;
+    return (methods.indexOf(upcased) > -1) ? upcased : method
   }
 
   function Request(url, options) {
@@ -266,6 +263,10 @@
       var xhr = new XMLHttpRequest()
       if (self.credentials === 'cors') {
         xhr.withCredentials = true;
+      }
+
+      if (methods.indexOf(self.method) === -1) {
+        reject(new TypeError('Method not allowed, must be one of ' + methods.join(', ')))
       }
 
       function responseURL() {
--- a/fetch.js
+++ b/fetch.js
@@ -212,7 +212,10 @@
 
   function normalizeMethod(method) {
     var upcased = method.toUpperCase()
-    return (methods.indexOf(upcased) > -1) ? upcased : method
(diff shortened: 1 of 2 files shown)


domenic:
fetch() shouldn't throw; it should reject. new Request should throw though.

(This is implicit in the spec. Since fetch() returns a Promise WebIDL type, it automatically catches thrown errors in the algorithm and turns them into rejections.)

mattking17:
I updated the patch, I think this is more in line with what you mentioned.

domenic:
Ideally `new Request({ method: 'BLADH' })` should throw a TypeError... The code in this polyfill seems structured rather different from the spec though.

mattking17:
Right...that's how it was originally in the patch, but maybe I misunderstood. If we throw on `new Request({ method: 'BLADH' })` then `window.fetch` polyfill will always throw when the `method` is bad, and will never get to the point of being able to reject.

I think it should just throw an error regardless of how it's called, instead of waiting for the request to be made.

domenic:
No, there's a differnece between `new Request` throwing, and `fetch` throwing. Check browsers.

mattking17:
This should cover throwing only `new Request...` but not when calling `window.fetch`

domenic:
Looks good :). It's a shame the structure here is so different from the spec (e.g. there being a Request.prototype.fetch method). That's what makes this complicated.

fetch.js
```diff
@@ -225,9 +225,14 @@
     this.mode = options.mode || null
     this.referrer = null
 
+    if (!options.constructedWithFetch && methods.indexOf(this.method) === -1) {
```
mislav: This consults `methods` array which is not a definite list of all supported HTTP methods, but only of few main methods that the spec defines as case-insensitive.

To properly implement this per-spec, we would have to maintain the [list of forbidden methods instead](https://fetch.spec.whatwg.org/#forbidden-method):

> A forbidden method is a method that is a byte-case-insensitive match for one of `CONNECT`, `TRACE`, and `TRACK`.

Furthermore, a ["method" is defined](https://tools.ietf.org/html/rfc7230#section-3.2.6) simply as a "token", which to me seems like it's basically any non-empty string.

So, we would have to check if the method is nonempty and if it's not listed among forbidden methods.

dgraham:
Closing as stale for now. Feel free to reopen if addressing @mislav's [review comments](https://github.com/github/fetch/pull/132#discussion_r49263515).

PR #140, opened 2015-05-12
More details around response URL

sandstrom:
- Browser support details.
- Clarify that query params are required.


mislav:
Thanks, but I don't think the old documentation was that unclear. After these changes, I don't think the docs around `responseURL` and `X-Request-URL` workaround are any better. :-1:

mislav:
This documentation was improved in 612d54ed7f2920c613d35271ffc4c6295d1bb0c4.

sandstrom:
@mislav Great! :boat:

PR #144, opened 2015-05-19
AMD and Commonjs support

ButuzGOL:
What do you think?

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,9 +1,20 @@
-(function() {
+;(function(root, factory) {
   'use strict';
 
-  if (self.fetch) {
+  if (root.fetch) {
     return
   }
+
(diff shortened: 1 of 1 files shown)


mislav:
This [breaks the build](https://travis-ci.org/github/fetch/jobs/63196946).

Also, we don't believe that a polyfill needs to be loaded in a modular fashion. Module loaders enhance isolation between modules and avoid polluting the global namespace. The whole point of this library is to "pollute" the global namespace by defining `window.fetch()`.

dgraham:
See https://github.com/github/fetch/pull/98 and https://github.com/github/fetch/issues/46.

ButuzGOL:
Thanks

PR #156, opened 2015-06-03
Properly convert undefined/null header values to strings.

kruppel:
In sending undefined/null header values with the fetch polyfill, I was getting an error:

```javascript
TypeError: 'undefined' is not an object (evaluating 'value.toString')

at normalizeValue (fetch.js:20:21)
    at Headers.append (fetch.js:42:14)
    at Headers.<anonymous> (fetch.js:30:15)
    at Array.forEach (native)
    at new Headers (fetch.js:29:45)
    at new Request (fetch.js:25:21)
    at self.fetch (fetch.js:273:18)
```

I tested this against the native window.fetch implementation in Chrome 43.0.2357.81 and observed that primitive values were handled fine when set as request header values. Using string concatenation for string conversion - over `toString` (as noted in [Axel Rauschmeyer's article on the subject](http://www.2ality.com/2012/03/converting-to-string.html)) - brings parity between the two implementations in this regard.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -7,7 +7,7 @@
 
   function normalizeName(name) {
     if (typeof name !== 'string') {
-      name = name.toString();
+      name = String(name)
     }
     if (/[^a-z0-9\-#$%&'*+.\^_`|~]/i.test(name)) {
       throw new TypeError('Invalid character in header field name')
@@ -17,7 +17,7 @@
 
   function normalizeValue(value) {
     if (typeof value !== 'string') {
-      value = value.toString();
+      value = String(value)
     }
     return value
   }
--- a/test/test.js
+++ b/test/test.js
@@ -128,7 +128,9 @@ suite('Headers', function() {
   test('converts field value to string on set and get', function() {
     var headers = new Headers()
     headers.set('Content-Type', 1)
+    headers.set('X-CSRF-Token', undefined);
     assert.equal(headers.get('Content-Type'), '1')
+    assert.equal(headers.get('X-CSRF-Token'), 'undefined')
   })
   test('throws TypeError on invalid character in field name', function() {
     assert.throws(function() { new Headers({'<Accept>': ['application/json']}) }, TypeError)


domenic:
String concatenation is actually incorrect as well. You should be doing `String(value)`.

String concatenation in particular gives the wrong result for `{ valueOf() { return 5; }, toString() { return "foo"; } }`.

kruppel:
> String concatenation in particular gives the wrong result for { valueOf() { return 5; }, toString() { return "foo"; } }.

Duly noted. Updated, thanks @domenic .

mislav:
Thanks for looking into this! The PhantomJS build is alright but SauceLabs in CI seems to be acting up and failing the build because it takes too long. You should pay no attention to that, but it might be useful to manually test in Firefox and IE if SauceLabs is unable to do it.

kruppel:
> You should pay no attention to that, but it might be useful to manually test in Firefox and IE if SauceLabs is unable to do it.

I ran the tests against my own Sauce Labs account. Here are the results:

```sh
⫸ SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="internet explorer" SAUCE_VERSION="11" make test
./node_modules/.bin/jshint *.js test/*.js
./script/test
Windows 7
internet explorer
11
{"suites"=>21, "tests"=>70, "end"=>nil, "passes"=>61, "reports"=>[], "start"=>nil, "duration"=>4394, "failures"=>0, "pending"=>9}
⫸ SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="internet explorer" SAUCE_VERSION="10" make test
./node_modules/.bin/jshint *.js test/*.js
./script/test
Windows 7
internet explorer
10
{"passes"=>61, "tests"=>70, "end"=>nil, "suites"=>21, "reports"=>[], "start"=>nil, "duration"=>2712, "failures"=>0, "pending"=>9}
⫸ SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="internet explorer" SAUCE_VERSION="9" make test
./node_modules/.bin/jshint *.js test/*.js
./script/test
Windows 7
internet explorer
9
{"suites"=>21, "tests"=>70, "end"=>nil, "passes"=>42, "reports"=>[], "start"=>nil, "duration"=>1364, "failures"=>0, "pending"=>28}
⫸ SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="firefox" SAUCE_VERSION="" make test
./node_modules/.bin/jshint *.js test/*.js
./script/test
Windows 7
firefox

"Test exceeded maximum duration after 180 seconds"
make: *** [test] Error 1
```

I gathered that a test is failing in Firefox for a reason unrelated to this change. Namely, Mocha's error handler by way of `window.onerror` is reporting a failure due to [this spec](https://github.com/github/fetch/blob/9e08a9f397ee89ba468ad693d83a199859f70d66/test/test.js#L467-474):

```sh
SyntaxError: JSON.parse: unexpected keyword at line 1 column 1 of the JSON data (:0)
process.on/global.onerror@http://localhost:3000/node_modules/mocha/mocha.js:6193:10
```

Global error handlers appear to handle errors thrown in promise fulfillment callbacks differently depending on vendor. Example:

```js
// http://jsbin.com/boqilipecu/1/edit?js,console

window.onerror = function(e) {
  console.error(e);
};

new Promise(function(resolve, reject) {
  resolve('{');
}).then(JSON.parse).catch(function() {});

// Firefox (38.0.5): logs an error
// Chrome (43.0.2357.81): nothing logged;
```

Again, completely separate, but if you're seeing flaky test failures against Firefox, it's likely due to the above. Otherwise, tests look to be passing.

mislav:
Thanks for the investigation. We'd welcome a separate PR that fixes Firefox on Sauce, if you know what's going on.

This looks ready to merge, although I'd like to figure out what's going on with Sauce first, as I don't want master to start failing again (although it's likely it would start failing on next build regardless of merging this). @josh do you have any idea why it's so abysmally slow that it eventually times out? Normal build times used to be under 2 minutes.

kolletzki:
:+1:

PR #157, opened 2015-06-05
Fix tests in Firefox.

kruppel:
1. ~~As mentioned in #156, the global error handler is catching errors thrown in promise fulfillment. To circumvent this, where faulty JSON parsing is expected, temporarily remove the mocha runner uncaught exception handler.~~

2. Properly passes an error to the rejection handler for `.json`. Resolving with `JSON.parse` (by reference) rather than invoking it in a new function results in the rejection callback receiving an `NS_ERROR_UNEXPECTED` exception (not an instance of an `Error`) rather than a `JSON.parse` `SyntaxError` instance.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -186,7 +186,9 @@
     }
 
     this.json = function() {
-      return this.text().then(JSON.parse)
+      return this.text().then(function(text) {
+        return JSON.parse(text)
+      })
     }
 
     return this
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
   matrix:
     - PHANTOMJS=1
     - SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="googlechrome" SAUCE_VERSION=""
+    - SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="firefox" SAUCE_VERSION=""
     - SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="internet explorer" SAUCE_VERSION="11"
     - SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="internet explorer" SAUCE_VERSION="10"
     - SAUCE_PLATFORM="Windows 7" SAUCE_BROWSER="internet explorer" SAUCE_VERSION="9"


test/test.js
```diff
@@ -443,6 +443,16 @@ suite('Body mixin', function() {
   })
 
   suite('json', function() {
+    if (mocha.runner) {
+      setup(function() {
+        runner.removeListener('uncaughtException', runner.uncaught)
```
mislav: Yikes, this means that during this suite, uncaught exceptions won't be handled my mocha as test failure right? Is there a way around this? Does the Firefox global error handler really capture exceptions in Promises even though they might be handled by its `catch` method?

test/test.js
```diff
@@ -465,7 +475,7 @@ suite('Body mixin', function() {
     })
 
     test('handles json parse error', function() {
-      return fetch('/json-error').then(function(response) {
```
mislav: I'm pretty sure we still need to return a Promise to Mocha, to keep the test running until we're done. Can't we return a resolved promise after we've handled the (expected) failure?

kruppel:
> 1. As mentioned in #156, the global error handler is catching errors thrown in promise fulfillment. To circumvent this, where faulty JSON parsing is expected, temporarily remove the mocha runner uncaught exception handler.

This actually isn't necessary.

> 2. Properly passes an error to the rejection handler for `.json`. Resolving with `JSON.parse` (by reference) rather than invoking it in a new function results in the rejection callback receiving an `NS_ERROR_UNEXPECTED` exception (not an instance of an `Error`) rather than a `JSON.parse` `SyntaxError` instance.

I'm now unable to reproduce the `NS_ERROR_UNEXPECTED` error, but you can view a trivial example (http://jsbin.com/gomiliyevu/1/edit?js,console) of how this issue surfaces.

test/.jshintrc
```diff
@@ -26,6 +26,7 @@
     "chai": false,
     "suite": false,
     "setup": false,
+    "teardown": false,
```
mislav: Is this necessary anymore?

mislav:
Thanks, I was able to reproduce the discrepancy in Firefox and confirm that your approach fixes it. To be honest, I'm not sure whether the Firefox way of implementing Promise callbacks is correct or not. Have you reported this to their team?

In any case, I'd like @josh @dgraham to take a :eyes: at this and then we can merge.

kruppel:
> Have you reported this to their team?

https://bugzilla.mozilla.org/show_bug.cgi?id=1172246

mislav:
Thanks for dealing with Bugzilla! You're a true hero :bow:

josh:
LGTM!

fetch.js
```diff
@@ -186,7 +186,9 @@
     }
 
     this.json = function() {
-      return this.text().then(JSON.parse)
+      return this.text().then(function(text) {
+        return JSON.parse(text)
+      })
```
dgraham: Is this a functional change, or just accommodates Firefox test behavior, or maybe both? It's disappointing to need a wrapper function just for tests.

domenic: Yeah it seems like it'd be better to report this Firefox bug on their bugtracker.

kruppel: Looks like there's movement in Bugzilla. If you want to just wait until the issue's fixed in Firefox, that's up to you. I honestly didn't anticipate the source of the problem being an outright browser bug when I opened the PR. Otherwise, I would've just started there. Thanks for the eyes in any case!

mislav: I'm :+1: for merging this change because it's innocent (adds an overhead of calling 1 more anon function) and fixes the current Firefox problem that might continue biting users that haven't upgraded to a future Firefox that may fix this.

dgraham:
Closing this in favor of the Bugzilla report of the Firefox bug. [Firefox 39](https://www.mozilla.org/en-US/firefox/39.0/releasenotes/) shipped with a native implementation of `fetch`, and the polyfill's usage there is quickly declining.

PR #164, opened 2015-06-15
Improve documentation for `fetch` caveats

mislav:
* Added caveat about cookies
* Revamped HTTP error handling
* Simpler `request.url` workaround explanation

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,27 +123,29 @@ ways that bear keeping in mind:
 
 #### Handling HTTP error statuses
 
-This causes `fetch` to behave like jQuery's `$.ajax` by rejecting the `Promise`
-on HTTP failure status codes like 404, 500, etc. The response `Promise` is
-resolved only on successful, 200 level, status codes.
+To have `fetch` Promise reject on HTTP error statuses, i.e. on any non-2xx
+status, define a custom response handler:
 
 ```javascript
-function status(response) {
+function checkStatus(response) {
   if (response.status >= 200 && response.status < 300) {
     return response
+  } else {
+    var error = new Error(response.statusText)
+    error.response = response
+    throw error
   }
-  throw new Error(response.statusText)
 }
 
-function json(response) {
+function parseJSON(response) {
   return response.json()
 }
 
 fetch('/users')
-  .then(status)
-  .then(json)
-  .then(function(json) {
-    console.log('request succeeded with json response', json)
+  .then(checkStatus)
+  .then(parseJSON)
+  .then(function(data) {
+    console.log('request succeeded with JSON response', data)
   }).catch(function(error) {
     console.log('request failed', error)
   })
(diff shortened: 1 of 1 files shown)


mislav:
@dgraham thoughts?

README.md
```diff
@@ -98,60 +98,90 @@ fetch('/users', {
...
+    console.log('request succeeded with JSON response', data)
   }).catch(function(error) {
     console.log('request failed', error)
   })
 ```
 
-### Response URL caveat
+#### Sending cookies
+
+To have `fetch` automatically use cookies for the current domain, the
```
dgraham: "To automatically send ..."

README.md
```diff
@@ -98,60 +98,90 @@ fetch('/users', {
...
+  even if the response is a HTTP 404 or 500. Instead, it will resolve normally,
+  and it will only reject on network failure, or if anything prevented the
+  request from completing.
+
+* By default, `fetch` **won't send any cookies** to the server, resulting in
+  unauthenticated requests if the site relies on maintaining a user session.
+
+#### Handling HTTP error statuses
+
+To have `fetch` Promise reject on HTTP error statuses, i.e. on any non-2xx
```
dgraham: "To reject the Promise on ..."

README.md
```diff
@@ -98,60 +98,90 @@ fetch('/users', {
...
 ```
 
-### Success and error handlers
+### Caveats
 
-This causes `fetch` to behave like jQuery's `$.ajax` by rejecting the `Promise`
-on HTTP failure status codes like 404, 500, etc. The response `Promise` is
-resolved only on successful, 200 level, status codes.
+The `fetch` specification differs from other ajax implementations in mainly two
+ways that bear keeping in mind:
```
dgraham: We need to mention jQuery specifically. It's where people migrating to fetch are coming from.

README.md
```diff
@@ -71,7 +71,7 @@ fetch('/users.json').then(function(response) {
 ```javascript
 var form = document.querySelector('form')
 
-fetch('/query', {
+fetch('/create', {
```
dgraham: `/users`

README.md
```diff
@@ -98,60 +98,90 @@ fetch('/users', {
...
+
+```javascript
+fetch('/users', {
+  credentials: 'same-origin'
+})
+```
+
+This option makes `fetch` behave similar to XMLHttpRequest with regards to
+cookies. Otherwise, cookies won't get sent, resulting in these requests not
+preserving the authentication session.
```
dgraham: It might be worthwhile to explain the `Set-Cookie` is a forbidden response header that is not accessible through `response.headers.get('Set-Cookie')`. We've answered that question a few times in issues.

README.md
mislav: You (the reader) are not the one who's rejecting the promise. Instead, you're receiving a promise that is awaiting resolution. Therefore, it doesn't seem correct to say "to reject the promise" doing this because you're not the actor here.

mislav:
Thanks for the review, @dgraham. Please take another :eyes:

mislav:
Thanks!

Issue #125, opened 2015-04-18
self is not defined


tcurdt:
I have an isomorphic react app. So while this the code is not running on node, still babel complains that "self is not defined". A quick hack that makes things work is to add "self = this;" to fetch.

Would be really nice to have a proper fix for this.

tcurdt:
just found https://github.com/matthew-andrews/isomorphic-fetch and then https://github.com/matthew-andrews/fetch/blob/master/fetch.js#L3 which has:

    if (typeof global !== 'undefined') {
      var self = global.self;
    }

mislav:
Yes please use that project instead. #31

tcurdt:
I don't want to really use it in isomorphic way - I just have an isomorphic setup. That's a bit of a different thing. (Which is why those projects are not really what I am after)

Ideally `require("whatwg-fetch")` would just export what it's offering.
What I ended up doing is:

    if (process.env.IS_BROWSER) {
      require('whatwg-fetch');
    }

Not nice - but works.

mislav:
Your workaround looks good. I would suggest applying the same conditional to all requires in your app that load libraries that are only useful in browsers.

tcurdt:
I will.

As a side note - I would much prefer this style of use

    var blafetch = require('whatwg-fetch');
    blafetch('/users.html')
      .then(function(response) {
        return response.text()
      }).then(function(body) {
        document.body.innerHTML = body
      })

or even better

    var Fetch = require('whatwg-fetch');
    Fetch.get('/users.html')
      .then(function(response) {
        return response.text()
      }).then(function(body) {
        document.body.innerHTML = body
      })

Maybe for the next major version.

mislav:
> As a side note - I would much prefer this style of use

Perhaps you're missing the point of this project. We're trying to create a polyfill for a future standard feature of browsers. The API for that feature is `window.fetch`. We're not trying to create anything more or other than that.

tcurdt:
I see.

billinghamj:
Unfortunately in React Native, the workarounds mentioned don't seem to be sufficient. Due to the way React packages JS files assuming they support AMD/CommonJS modularization properly, this library ends up running anyway - even if it isn't required.

mislav:
@billinghamj We welcome a PR that solves this while keeping in mind that the code must also be able to run in a web worker environment. See https://github.com/github/fetch/pull/239#issuecomment-160598283 for context

billinghamj:
[...]

PR #165, opened 2015-06-17
Bail if `self` does not exist

c0:
- Exit out of attempting to polyfill when `self` does not exist.

Context: We're prerendering React components on the server. While `fetch` won't actually be executed, it throws an error because it's looking for `self`, which does not exist in the server environment.

This is not an attempt to make it run in node, but to make it return without throwing an error.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,7 +1,7 @@
 (function() {
   'use strict';
 
-  if (self.fetch) {
+  if (typeof self === 'undefined' || self.fetch) {
     return
   }
 


mislav:
Please use the workaround suggested in https://github.com/github/fetch/issues/125#issuecomment-94266403

c0:
The workaround doesn't apply with ES6's `import` (because it needs to be at the top level).

mislav:
Is there any way you could prevent specific client-side JS libraries to load on the server?

dgraham:
Closing as stale.

PR #179, opened 2015-07-19
Allow passing a Request instance to Request constructor

### LINKED ISSUE #177: Incorrect Request implementation
Operating system: OS X Yosemite Version 10.10.4
Browser: Safari Version 8.0.7 (10600.7.12)

I've been developing with the following browsers:
* Firefox Developer Edition 41.0a2 (2015-07-14)
* Google Chrome Canary Version 46.0.2456.0 canary (64-bit)

Both Firefox and Chrome implement the expected behavior.

Code sample:

```js
var request = new Request('/foo')
console.log(request.url)
// Chrome: "https://babeljs.io/foo"
// Safari: "/foo"

var request2 = new Request(request)
console.log(request2.url)
// Chrome: "https://babeljs.io/foo"
// Safari: {"url":"/foo","credentials":"omit","headers":{"map":{}},"method":"GET","mode":null,"referrer":null,"_bodyText":""}
```

The implementation in this repository is always taking the first parameter and setting it as the URL, but this is inconsistent with the [standard](https://fetch.spec.whatwg.org/#request-class), which allows you to pass a Request instance as the first parameter.

In the short-term, please consider adding a note to the README to inform users of this inconsistency. Tracking down this issue in Safari proved to be quite tricky for my project.

mislav:
I tried to match the spec as closely as possible. This should also make it possible to use `fetch(request, options)` invocation.

Fixes #177

/cc @dgraham @cesarandreu

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -209,7 +209,9 @@
       }
       this.url = input.url
       this.credentials = input.credentials
-      if (!options.headers) this.headers = new Headers(input.headers)
+      if (!options.headers) {
+        this.headers = new Headers(input.headers)
+      }
       this.method = input.method
       this.mode = input.mode
       if (!body) {
@@ -221,7 +223,9 @@
     }
 
     this.credentials = options.credentials || this.credentials || 'omit'
-    if (options.headers || !this.headers) this.headers = new Headers(options.headers)
+    if (options.headers || !this.headers) {
+      this.headers = new Headers(options.headers)
+    }
     this.method = normalizeMethod(options.method || this.method || 'GET')
     this.mode = options.mode || this.mode || null
     this.referrer = null
--- a/fetch.js
+++ b/fetch.js
@@ -200,20 +200,36 @@
     return (methods.indexOf(upcased) > -1) ? upcased : method
   }
 
-  function Request(url, options) {
+  function Request(input, options) {
     options = options || {}
-    this.url = url
+    var body = options.body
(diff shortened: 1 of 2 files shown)


cesarandreu:
Oh awesome! That's was really fast.

cesarandreu:
I just looked through the spec and comparing it with Chrome Canary they seem to at least behave the same.

mislav:
We got a legitimate failure in Chrome because it doesn't conform to the spec fully: a Request with an already consumed body should throw an exception, and in Chrome it doesn't. @dgraham What do we do in these situations?

dgraham:
@mislav I usually test the behavior in Firefox as well, to determine if both browsers agree on the implementation. It could be a misreading of the specification or a real Chrome bug.

mislav:
It passes in Firefox, as well as my polyfill implementation based on the spec. It's not a super-important feature, though— it just throws a TypeError early when the body has already been consumed. It looks like Chrome implementation in general allows bodies to be re-consumed.

cesarandreu:
Possibly related to the following?
https://github.com/whatwg/fetch/issues/55

mislav:
Thanks @cesarandreu.

@dgraham It's a real Chrome bug due to the spec only formalizing this very recently. Should I skip the test in Chrome?

dgraham:
@mislav That sounds fine. We can re-enable the test after Chrome ships a fix for the bug.

PR #188, opened 2015-08-05
Body.json with reviver

heruan:
`Body.json()` should take an optional parameter to be passed to `JSON.parse(json[, reviver])` function to use a _reviver_ callback to parse JSON data.

P.S. I understand this is a polyfill, see https://github.com/whatwg/fetch/issues/104

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -186,6 +186,12 @@
     }
 
     this.json = function() {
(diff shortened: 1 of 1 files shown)


heruan:
Such feature will not be present in the specs.

PR #189, opened 2015-08-08
Update bower.json

camsong:
Update es6-promise version


mislav:
This fails the build. It doesn't seem that the 2.3 version is available?

dgraham:
Duplicate of #191.

PR #191, opened 2015-08-11
Update es6-promise version

camsong:


mislav:
Why are we doing this? Won't this force everyone who updates their fetch.js via Bower to upgrade es6-promise version even if they wanted to stay frozen at v1.0?

Our library is compatible with es6-promise 1.0 so I don't see the reason for this bump, except testing forward compatibility in our CI.

camsong:
Just want to keep packages up to date. Close for now.

Issue #184, opened 2015-07-30
Node.js support


kirill-konshin:
Do you have anything in your roadmap regarding Node.js support?

The purpose is to allow other cross-env libraries to use `fetch` as the client.

From what I see now in `master` there are couple of things to be done:

1. Use `XMLHttpRequest` polyfill (for example https://github.com/pwnall/node-xhr2)
2. Wrap the code in something like UMD to make sure CommonJS and AMD compatibility
3. Export all classes and `fetch()` function instead of always writing to global
4. Add optional `polyfill()` method just like in `es6-promise` for AMD users
5. Take over NPM package name `fetch`, current one is unmaintained for 2 years

Is it worth investigation and making a PR? Does not seem to be too much effort.

matthew-andrews:
@kirill-konshin have you seen the [`node-fetch` project](https://github.com/bitinn/node-fetch).  If you want a single library that will pull both through you might even like to [consider using mine as well](https://github.com/matthew-andrews/isomorphic-fetch).

kirill-konshin:
Github's polyfill is the most popular and well known, so I thought maybe it is reasonable to make it isomorphic...

I personally prefer single library approach so that I won't have to deal with bugs in 2-3 libraries which potentially may not catch up with the spec/bleeding edge implementations :)

Your version is not UMD-wrapped (will not work with RequireJS at all) and it is not a one-file dependency as this repo (yours has own sub-dependencies, which means slightly trickier install for non-AMD users and front end developers in general).

As I said earlier, those 5 items does not seem to be hard to do.

matthew-andrews:
You make some really good points that I hadn't thought about before because I don't really use RequireJS.

I remember point 5 was considered but rejected because in order not to break backwards compatibility this repo would have to jump to v2.0.0… which didn't make sense…  Once you know it's `whatwg-fetch` though it's not too difficult to find, is it?

I can't speak for the GitHub team but much of what you're asking for seems to take this project away from what I think it is defined as:- a client side implementation of the whatwg fetch specification and nothing more.  Implementing specific 

I don't think it would be very difficult to make a little requirejs-isomorphic-fetch wrapper similar to my one (perhaps mine ought to have been called `commonjs-isomorphic-fetch`).  Maybe you could do it? :)

[...]

PR #192, opened 2015-08-12
fetch is now available as a polyfill or a self-contained library

steveluscher:
This refactor permits usage of whatwg-fetch as:

* a polyfill that modifies `self`

  `require('whatwg-fetch');`

* or as a self-contained module that doesn't

  `var fetch = require('whatwg-fetch/library');`

Should you use the second form in a project whose code is sometimes executed on the server and sometimes on the client (as is the case with certain classes of apps that do server rendering, or are tested in headless environments), this has the nice side effect of not throwing an exception on import just because `self` is undefined.

This is similar to the approach taken by [core-js](https://github.com/zloirock/core-js#core-js).


mislav:
Sorry, we're not fans. https://github.com/github/fetch/issues/184#issuecomment-126774267 https://github.com/github/fetch/pull/31#issuecomment-62178920

If we accept this pull request, we can't use `fetch` on GitHub.com anymore, where we don't use any JS module loader. It would be pretty hard to convince us to make a change that prevents us from using our own library in our own product.

steveluscher:
Thanks for taking a look!

I think you *could* continue to use it on GitHub, though. This PR builds a `./fetch.js` that works just like your old one – with no module loader needed. It's only the additional (and optional) `./library/index.js` version that follows the CommonJS pattern.

mislav:
> This PR builds a `./fetch.js` that works just like your old one – with no module loader needed.

The creation of `./fetch.js` seems to require an extra build step with webpack. That adds complexity to our assets setup. Right now we simply fetch each JS library we use with Bower and load it directly into the site, no post-processing or intermediate build steps needed.

Your "compiled" version of `fetch.js` is also 0.5kB larger than ours after minification. Not a huge deal, but one more reason against our adoption of that approach.

Thanks for taking the time to break down our polyfill into modules, though. It's a good example of how it could be done if a library wanted to take that approach. We feel that our polyfill is still simple enough that it can stay in a single file.

PR #196, opened 2015-08-18
UMD compatibility

lagden:


mislav:
Sorry; here's how we feel about either isomorphic support or module loader compatibility: https://github.com/github/fetch/issues/184#issuecomment-126774267

lagden:
But UMD works everywhere!!! :cry:
Well... I tried!

mislav:
[Many have tried](https://github.com/github/fetch/issues?utf8=%E2%9C%93&q=module+is%3Aunmerged+).

PR #205, opened 2015-09-28
Fixing the `credentials` option documentation

cloderic:
According the the code, 'same-origin' is not used but 'include' results in the underlying `XMLHttpRequest`'s `withCredentials` being set to _true_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ must be provided:
 
 ```javascript
 fetch('/users', {
-  credentials: 'same-origin'
+  credentials: 'include'
 })
 ```
 


cloderic:
The concerned code is at https://github.com/github/fetch/blob/master/fetch.js#L333-L335

mislav:
I can see how this could have been confusing.

Yes, `same-origin` isn't used in the polyfill. This is because the polyfill is made using XHR which sends cookies to same domains by default (no opt-in needed). This behavior in native fetch is opt-in using `same-origin`, so the README suggests you use this so your code works on modern Chrome and Firefox which implement fetch.

Now, `include` is also a valid option value, but also instructs `fetch` to send cookies and authentication info to other domains too. This requires special opt-in from XHR as well, which you noticed in the code. However, dealing with CORS is not something we wanted to cover in the README documentation. We just wanted to cover the simplest case of sending cookies for same-domain requests. So we did, with `same-origin`.

cloderic:
Is there at least a generally accepted documentation for `fetch` that can be referenced?

I was switching from [request](https://github.com/request/request) to `fetch` and got stuck with cors usage. I had to dive into the code to understand what to do exactly, not the perfect DX if you ask me...

mislav:
Yeah we understand that the spec might not be the prettiest documentation to read if you're just trying to get work done. I'm not opposed to maintaining some usage documentation for fetch in general in this repo. We already documented the common caveats in the README.

If @dgraham agrees, you could open another issue in this tracker with a list of things that you'd like to see documented. Or, submit some documentation that we might adopt as aa starting point for this project.

PR #208, opened 2015-09-29
Improve error messaging in case of failed network request

jbaumbach:
Might save beginners quite a bit of head-scratching time when they try to POST w/out setting the header to the correct content type.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -311,7 +311,13 @@
       xhr.onload = function() {
         var status = (xhr.status === 1223) ? 204 : xhr.status
         if (status < 100 || status > 599) {
-          reject(new TypeError('Network request failed'))
+          var msg = 'Network request failed'
+          if (xhr.responseText) {
+            msg += ' ("' + xhr.responseText + '")'
+          } else {
+            msg += ' (no other details)'
+          }
+          reject(new TypeError(msg))
           return
         }
         var options = {


jbaumbach:
FYI: test seem to run ok locally - maybe something is wrong with Travis CI config?

fetch.js
```diff
@@ -311,7 +311,13 @@
       xhr.onload = function() {
         var status = (xhr.status === 1223) ? 204 : xhr.status
         if (status < 100 || status > 599) {
-          reject(new TypeError('Network request failed'))
+          var msg = 'Network request failed'
+          if (xhr.responseText) {
+            msg += ' ("' + xhr.responseText + '")'
+          } else {
+            msg += ' (no other details)'
```
mislav: not a fan of "no other details"

fetch.js
```diff
@@ -311,7 +311,13 @@
       xhr.onload = function() {
         var status = (xhr.status === 1223) ? 204 : xhr.status
         if (status < 100 || status > 599) {
-          reject(new TypeError('Network request failed'))
+          var msg = 'Network request failed'
+          if (xhr.responseText) {
+            msg += ' ("' + xhr.responseText + '")'
```
mislav: What will responseText typically contain exactly? Which browsers did you test this on? Would it be possible to create automated tests? What does native fetch return in Chrome and Firefox?

fetch.js
```diff
@@ -311,7 +311,13 @@
       xhr.onload = function() {
         var status = (xhr.status === 1223) ? 204 : xhr.status
         if (status < 100 || status > 599) {
```
dgraham: This check is in place only for IE 9, which can set `xhr.status` to invalid values. Unless another browser is reaching this block unintentionally, I think we can close this pull request.

jbaumbach: False. This happens with React Native. The specific instance is if the
method is POST and the content type is not specified. The status is "0" and
that field is filled in with the error message. Without the description
included here, the client has no idea what the error is.

fetch.js
jbaumbach: That part is not as important as providing the other info. I could update the code to omit that else part (although it's nice to let the user know that info as well).

fetch.js
dgraham: The `onload` event must be fired when "[the request has successfully completed](http://www.w3.org/TR/XMLHttpRequest/#event-xhr-load)", which means there is an HTTP status code available in the response from the server. If there is no response, as in the case of a network failure, `onerror` must be triggered.

Here's [the commit](https://github.com/github/fetch/commit/5a2d5f860b399f6d4dcac302788f77d594b6adc1) describing the current status check for IE 9.

I'm not familiar with React Native, but this might be a bug in the implementation of XHR. The [supported browsers](https://github.com/github/fetch#browser-support) are listed in the readme.

Issue #185, opened 2015-07-31
Support cloning of Responses


kirill-konshin:
According to MDN https://developer.mozilla.org/en-US/docs/Web/API/Response/clone and the spec https://fetch.spec.whatwg.org/#dom-response-clone `Response` class should have a `clone()` method, which seems to be missing now.

mislav:
Good catch. Care to make a pull request?

kirill-konshin:
Cloning of `Headers` is trivial -- just call `new Headers(this.headers)`, primitives can be easily cloned as well, the biggest question is how to clone body's Blob or FormData instances and whether it is needed.

It seems that `Blob` can be cloned by calling `blob.slice(blob.size)` which returns a new blob. But `readBlobAs*` functions in polyfill can read same blob multiple times and since blobs are immutable, it should be safe to just pass the same blob as new body.

From what I see in https://developer.mozilla.org/en-US/docs/Web/API/FormData there is no way to get keys/values and iterate through them filling a new FormData instance with values... But based on implementation nuances and lack of ability to actually read `FormData` I think it's fine to pass the same instance of FormData body too.

The new `Response` will have `bodyUsed = false` anyway, so even if source `Response` has been read already, new one will still be readable...

If you agree with passing non-cloned bodies then I can make a PR in the next few days.

P.S. Method `Body.formData()` should resolve with a `FormData` object, not with a text:

```js
this.formData = function() {
    return this.text().then(decode)
}
```

In polyfill `text()` with `FormData` body will throw an exception anyway. I should probably file a separate bug for that.

mislav:
> Method `Body.formData()` should resolve with a FormData object, not with a text:

Thanks for looking into this. One thing that baffles me is that Chrome's fetch implementation doesn't have the `formData()` method for some reason.

kirill-konshin:
I suppose that `Body.formData()` is not implemented because `FormData` is totally useless so far. You can write but you cannot read.

I'll make a PR.

kirill-konshin:
After quite a while I finally had time to implement PR: https://github.com/github/fetch/pull/209

PR #209, opened 2015-09-29
Fix for #185

kirill-konshin:


test/test.js
```diff
@@ -376,6 +376,36 @@ suite('Response', function() {
       return json;
     })
   })
+  test('clone text response', function() {
+    var r = new Response('{"foo":"bar"}', {headers: {'content-type': 'application/json'}}),
+        clone = r.clone();
+    assert(clone.headers !== r.headers, true);
+    assert(clone.headers.get('content-type') === r.headers.get('content-type'), true);
+    return Promise.all([clone.json(), r.json()]).then(function(jsons){
+      assert(jsons[0].foo, 'bar');
```
mislav: I think you wanted to use `assert.equal` here. Also, to compare two hashes which are supposed to be the same, you could use `assert.deepEqual(jsons[0], jsons[1])`

test/test.js
```diff
@@ -376,6 +376,36 @@ suite('Response', function() {
...
+      assert(jsons[1].foo, 'bar');
+      return jsons;
```
mislav: This final `return` probably isn't needed

test/test.js
```diff
@@ -376,6 +376,36 @@ suite('Response', function() {
       return json;
     })
   })
+  test('clone text response', function() {
+    var r = new Response('{"foo":"bar"}', {headers: {'content-type': 'application/json'}}),
+        clone = r.clone();
+    assert(clone.headers !== r.headers, true);
```
mislav: There's no need to pass `true` as 2nd argument to `assert`. The 2nd argument is optional message in case of failure

test/test.js
kirill-konshin: Thanks, I'm more used to expect(...).to... syntax. Give me a sec, I'll upload update.

test/test.js
kirill-konshin: I copy-pasted that from another test. I'll remove that.

test/test.js
```diff
@@ -376,6 +376,36 @@ suite('Response', function() {
...
+  test('clone blob response', function() {
+    return fetch('/binary').then(function(response) {
+      return Promise.all([response.clone().arrayBuffer(), response.arrayBuffer()]).then(function(bufs){
+        assert(bufs[0] instanceof ArrayBuffer, 'buf is an ArrayBuffer instance')
+        assert.equal(bufs[0].byteLength, 256, 'buf.byteLength is correct')
+        var view0 = new Uint8Array(bufs[0])
+        for (var i = 0; i < 256; i++) {
+          assert.equal(view0[i], i)
+        }
+        assert(bufs[1] instanceof ArrayBuffer, 'buf is an ArrayBuffer instance')
```
mislav: You're repeating identical logic as for `bufs[0]` here. How about iterating through both `bufs` in a loop and applying the same block of assertions?

mislav:
It's also necessary to manually run these tests against native `fetch` implementation in latest Chrome and Firefox (although Chrome might be covered by Saucelabs in CI, unless that's busted, which it often is). Please confirm when you've done that; thanks!

kirill-konshin:
I've uploaded all changes based on your comments. Yes, I confirm, that I've run tests for both native and polyfill. To do that I had to insert a line `window.fetch = null`. I suggest to create a second `test.html` with that specific case.

Also I had to skip the test of arrayBuffer under Phantom... But it's what is usually done in your tests :)

Some test runs are failing on Travis:

```
["Windows 7", "googlechrome", ""]
"Invalid tunnel_identifier 378.2"
```

How to fix this?

mislav:
The "Invalid tunnel_identifier" error is know to us and we're working on fixing it. It's not your fault.

I've merged your changes, thanks. I've got some feedback for you to keep in mind for your future contributins to projects:

* We like to see more descriptive pull request titles and messages. And instead of putting `Fix for #185` in the title, put `Fixes #185` in the pull request message so that the referenced issue gets auto-closed when the PR is merged.
* We like to see more descriptive commit messages. Your commit was simply `Fix for #185`, but when listing commits on the command-line with `git log`, that commit title is not very friendly.
* I had to amend your commit to fit our coding style. For instance, we don't use semicolons. When adding code, always look what the coding style in the rest of the codebase is like.
* The `Request.clone()` method was broken (it didn't copy the body properly) and we didn't catch that because it didn't have tests. I've added tests and fixed the method. fd362dd

kirill-konshin:
Thanks for your recommendations.

PR #212, opened 2015-10-05
Support Response.error() and Response.redirect()

bryanrsmith:
This change adds support for the [Response.error()](https://fetch.spec.whatwg.org/#dom-response-error) and [Response.redirect()](https://fetch.spec.whatwg.org/#dom-response-redirect) static methods. Firefox and Chrome both support these methods.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -280,6 +280,22 @@
 
   Body.call(Response.prototype)
 
+  Response.error = function() {
+    var response = new Response(null, {status: 0, statusText: ''})
+    response.type = 'error'
+    return response
+  }
+
+  var redirectStatuses = [301, 302, 303, 307, 308]
+
+  Response.redirect = function(url, status) {
+    if (redirectStatuses.indexOf(status) === -1) {
+      throw new RangeError('Invalid status code')
+    }
+
+    return new Response(null, {status: status, headers: {location: url}})
+  }
+
   self.Headers = Headers;
   self.Request = Request;
   self.Response = Response;
--- a/test/test.js
+++ b/test/test.js
@@ -376,6 +376,21 @@ suite('Response', function() {
       return json;
     })
   })
+
+  test('error creates error Response', function() {
+    var r = Response.error()
+    assert(r instanceof Response)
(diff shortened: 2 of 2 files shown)


fetch.js
```diff
@@ -280,6 +280,22 @@
...
+    var response = new Response(null, {status: 0, statusText: ''})
+    response.type = 'error'
+    return response
+  }
+
+  var redirectStatuses = [301, 302, 303, 307, 308]
+
+  Response.redirect = function(url, status) {
+    if (redirectStatuses.indexOf(status) === -1) {
+      throw new RangeError('Invalid status code')
```
mislav: Is RangeError defined in all the browsers that we currently support?

test/test.js
```diff
@@ -376,6 +376,21 @@ suite('Response', function() {
...
+    assert.equal(r.status, 0)
+    assert.equal(r.statusText, '')
+    assert.equal(r.type, 'error')
+  })
+
+  test('redirect creates redirect Response', function() {
+    var r = Response.redirect('/hello', 301)
+    assert(r instanceof Response);
+    assert.equal(r.status, 301)
+    assert.equal(r.headers.get('Location', '/hello'), '/hello')
```
mislav: Did you run these tests in browsers that implement fetch such as Chrome and Firefox?

fetch.js
bryanrsmith: Yep! Added in ES3.

test/test.js
bryanrsmith: Yes, I forgot to mention that. They pass in Chrome, Firefox, and Safari. I did not manually test IE/Edge or any mobile browsers. I also manually inspected the `Response` objects created by Chrome and Firefox to verify that the instances created here look similar in terms of data not mentioned by the spec.

bryanrsmith:
No rush, but please let me know if there's any more I can do on this.

mislav:
@bryanrsmith Sure; thanks for the comments!

@dgraham This looks good to you?

dgraham:
:shipit:

PR #213, opened 2015-10-05
Remove superfluous assignment

bryanrsmith:
`this.url` is unconditionally assigned at the end of the constructor, so this assignment is unnecessary.


mislav:
Good catch!

PR #216, opened 2015-10-09
Fix incorrect assertion

bryanrsmith:
`assert()` checks for a truthy first parameter, and uses the second as the error message.


mislav:
Good catch!

PR #217, opened 2015-10-09
Reject the Promise returned by fetch() when Request ctor throws

bryanrsmith:
This solves part of what #132 does, but that PR seems abandoned.

https://fetch.spec.whatwg.org/#dom-global-fetch step 2 states that exceptions thrown by the Request constructor should cause fetch() to reject with the error. Exceptions thrown synchronously in the Promise callback will cause a rejection, so I've simply moved the Request ctor call into the Promise callback.

Manually verified that tests pass with native fetch implementations in Chrome and Firefox.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -55,6 +55,14 @@ test.skip('rejects promise for network error', function() {
   })
 })
 
+test('rejects when Request constructor throws', function() {
+  return fetch('/request', { method: 'GET', body: 'invalid' }).then(function() {
+    assert(false, 'Invalid Request init was accepted')
+  }).catch(function(error) {
+    assert(error instanceof TypeError, 'Rejected with Error')
+  })
+})
+
 // https://fetch.spec.whatwg.org/#headers-class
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -601,8 +609,7 @@ suite('Methods', function() {
     })
   })
 
-  // TODO: Waiting to verify behavior
```
bryanrsmith: Is this change OK? I'm not sure what this TODO was for, but as far as I can tell the tests appear correct and inline with the spec and current implementations. I needed to enable them to verify that the Request constructor throws under the same conditions I'm using to verify that fetch() rejects.

mislav: Yeah, right now the Chrome implementation also throws an exception on GET/HEAD with body. There is a [suggestion to change the spec to allow](https://github.com/whatwg/fetch/issues/83) bodies for GET requests, but until the spec actually changes, I think these tests are goo.d

mislav:
Looks good!

PR #218, opened 2015-10-11
Expand URL in `Reponse.redirect()`

mislav:
Native Chrome and Firefox implementations seem to expand the URL to include scheme, host, and port.

https://travis-ci.org/github/fetch/jobs/84727155

References #212 /cc @bryanrsmith

@dgraham: I didn't notice the Sauce failure because they're intermittently failing more often than actually running :/ Only after merging above PR to master did the Chrome test actually run and reveal a bug.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -292,6 +292,10 @@
       throw new RangeError('Invalid status code')
     }
 
+    var link = document.createElement('a')
+    link.href = url
+    url = link.href.toString()
+
     return new Response(null, {status: status, headers: {location: url}})
   }
 
--- a/test/test.js
+++ b/test/test.js
@@ -395,9 +395,10 @@ suite('Response', function() {
 
   test('redirect creates redirect Response', function() {
     var r = Response.redirect('/hello', 301)
+    var expectedUrl = location.protocol + '//' + location.host + '/hello'
     assert(r instanceof Response);
     assert.equal(r.status, 301)
-    assert.equal(r.headers.get('Location', '/hello'), '/hello')
+    assert.deepEqual(r.headers.getAll('Location'), [expectedUrl])
   })
 })
 


fetch.js
```diff
@@ -292,6 +292,10 @@
       throw new RangeError('Invalid status code')
     }
 
+    var link = document.createElement('a')
+    link.href = url
```
mislav: Is there a better way to do this? Would there be downsides to intializing `A` elements?

fetch.js
```diff
@@ -292,6 +292,10 @@
       throw new RangeError('Invalid status code')
     }
 
+    var link = document.createElement('a')
```
mislav: This is failing inside service worker because `document` is unavailable. Any suggestions? https://travis-ci.org/github/fetch/jobs/84728696

bryanrsmith:
I'm not sure how I missed the failing test in Chrome. Sorry about that @mislav!

It looks like this is also an issue with the `Request` constructor, which should also expand the `url`.

Web workers do have access to `location`, so we could manually build the url.

mislav:
> Web workers do have access to location, so we could manually build the url.

Manually expanding the URL isn't so trivial as it seems. We need to support relative paths (`../foo/bar`), URLs that specify the hostname but reuse the scheme (`//example.com/foo`), and finally URLs that are absolute and override all other components (`http://example.com/foo`). Without bringing in URL polyfill and without `document.createElement()` support, I don't know how to approach this.

bryanrsmith:
It sounds like the options are:

1. Polyfill URL
2. Manually expand the URL
3. Don't expand the URL and be OK with the inconsistency

The URL polyfill is pretty large, and it doesn't seem great for users to require that.

I know it's not entirely trivial to do manually, but it seems like the best option to me. I'd be happy to work on it if you agree.

bryanrsmith:
A little more info:

* According to the spec, `Response.redirect`'s `url`, and the `Request` ctor's `input` and `init.referrer` should all be resolved and expanded. (This polyfill currently always sets `referrer` to `null`, though.)
* `document.baseURI` would need to be accounted for in manual resolution code

I've opened #219 to let the `Response.redirect` test behave consistently when run against native implementations vs the polyfill until a decision is reached on this issue. 

I think the reason I missed the test failure in Chrome was that I originally used an absolute URL, but changed it before committing :-/

mislav:
Closing due to technical difficulties with service workers, and the fact that implementing this isn't terribly important. If the need arises, we'll revisit this.

PR #227, opened 2015-10-30
Allow making a POST request with an ArrayBuffer body

joshuawarner32:
This enables POSTing with an ArrayBuffer body, but doesn't change the existing support for receiving ArrayBuffer responses (via blob)

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -116,7 +116,8 @@
         return false
       }
     })(),
-    formData: 'FormData' in self
+    formData: 'FormData' in self,
+    arrayBuffer: 'ArrayBuffer' in self
   }
 
   function Body() {
@@ -133,6 +134,9 @@
         this._bodyFormData = body
       } else if (!body) {
         this._bodyText = ''
+      } else if (support.arrayBuffer && ArrayBuffer.prototype.isPrototypeOf(body)) {
+        // Only support ArrayBuffers for POST method.
+        // Receiving ArrayBuffers happens via Blobs, instead.
       } else {
         throw new Error('unsupported BodyInit type')
       }
--- a/test/test.js
+++ b/test/test.js
@@ -205,6 +205,27 @@ suite('Request', function() {
     })
   })
 
+  ;(ArrayBuffer in self ? test : test.skip)('sends ArrayBuffer body', function() {
+    var text = 'name=Hubot'
+
+    var buf = new ArrayBuffer(text.length)
+    var view = new Uint8Array(buf)
+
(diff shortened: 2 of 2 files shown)


joshuawarner32:
The travis tests are failing with this error, which I'm not able to reproduce locally, but I don't think it relates to my changes.  Any tips are appreciated.

```
./script/test
....https://saucelabs.com/jobs/f9fc0d513cc9426690b8a03b710a27b2
["Windows 7", "internet explorer", "9"]
"Invalid tunnel_identifier 402.5"
make: *** [test] Error 1
```

fetch.js
```diff
@@ -363,7 +367,7 @@
       }
 
       if ('responseType' in xhr && support.blob) {
-        xhr.responseType = 'blob'
+        xhr.responseType = 'arrayBuffer'
```
mislav: Can you explain why was this particular change needed?

mislav:
I'm sorry for the CI failure, it wasn't your fault. I just updated master with an attempted fix for CI. Can you merge master into your branch and push to update this PR? It should get the CI passing. Thanks

fetch.js
joshuawarner32: Whoops!  Good catch.  That was some code I had in to debug what turned out to be an unrelated issue.  I've since removed this change.

joshuawarner32:
Ok, I rebased onto master, but it looks like one of the jobs is still failing.  Is there a way I can see which tests are failing?  I'm not seeing it in the build output on travis.

mislav:
It [failed for IE 9](https://saucelabs.com/jobs/e1a5e67072e24613a9a6f4816aea6c26).

You'll probably have to skip your added test on browsers that don't support ArrayBuffer.

joshuawarner32:
@mislav - Looks like the tests passed this time.  Are there any other concerns you have?

mislav:
Thanks!

joshuawarner32:
Would it be possible to get a new npm release with this code?  If now's not the right time, that's fine - we can use a full git url for the time being.

mislav:
I've pushed a tag right now. Should appear on npm automatically.

joshuawarner32:
Hmm, I think something may have gone wrong with that automated release process.  I don't see the new release on npm yet.  Or, perhaps, it's just very slow; feel free to say "be patient!".

mislav:
Hm something went wrong with auto-publishing to npm from Travis CI. I published manually now.

joshuawarner32:
Looks good, thanks!

PR #230, opened 2015-11-11
README: error block updates

urakozz:
checkStatus should return Promise


dgraham:
There's no need to wrap return values or exceptions in Promise objects.

http://www.html5rocks.com/en/tutorials/es6/promises/

Issue #231, opened 2015-11-15
Using this library in commonjs environment


IonicaBizau:
This library is :cool:!!! I'm using it in one of my projects and I want to `require` it in my code, bundling everything with Browserify.

The `fetch.js` file doesn't `module.exports` anything currently. I would suggest to make it export the `fetch` function (even it's the native one or not).

Then we could do:

```js
var fetch = require("whatwg-fetch")

fetch(...)
```

Just an idea. I'm not sure if it's good to be implemented or not since the `fetch` function is anyway a global, but the code looks better and clearer this way than just `require("whatwg-fetch")` (without a variable name) in my file. :memo: 

Looking forward for your feedback. :smile:

dgraham:
See https://github.com/github/fetch/issues/184#issuecomment-126774267 for a discussion of module loaders.

IonicaBizau:
@dgraham Ohhh, totally understand! Thanks! :innocent:

Issue #47, opened 2014-12-24
Better error when trying to `.json` a `Resource` that's not JSON.


matthew-andrews:
It would be nice if there were a better error when trying to `.json` a `Resource` that's not JSON.

Currently in two implementations ([this polyfill](https://github.com/github/fetch) and Chrome Service Workers) trying to `fetch` and `.json()` a non JSON resource throws a not particularly helpful error:-

```js
fetch("//some.thing/not-a-json-resource")
  .then(function(res) {
    return res.json();
  })
  .catch(function(err) {
    console.log(err); // => Unexpected token c
  });
```

![screen shot 2014-12-24 at 12 36 59](https://cloud.githubusercontent.com/assets/825088/5548458/961311ce-8b69-11e4-8824-4ebb8bf7423f.png)

It would make for more less unpleasant debugging / error aggregation if it threw an error along the lines of “Failed to parse {{url}} as JSON”… or something like that…

[I know errors aren't spec'd](https://github.com/github/fetch/pull/44#issuecomment-67579609) as such so it's not something that should/could be fixed in the spec — but I will raise as an issue in Chrome as well so that GitHub & Chrome's fetches can stay aligned but I think it's something that should be improved and I'm sure you have an idea of what it might look like…

josh:
Have any specific ideas for how we could improve this? Or should we just wait it out until Chrome makes the first move?

dgraham:
I think this error is pretty good. "SyntaxError" and "unexpected token" are good clues that's it's invalid JSON. Then you can inspect the response and see it's HTML or other non-JSON data.

matthew-andrews:
for what it's worth i'm playing with ideas here:-
https://github.com/matthew-andrews/fetchres#usage-node

PR #239, opened 2015-11-29
Renamed `self` to `window` since we're in a browser environment and want to support jsdom/jsdomify

### LINKED ISSUE #238: rename self to window
otherwise we can't have tests that run on a node runtime environment simulating a browser (e.g. using jsdom or jsdomify)

niftylettuce:
Closes #238

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,7 +1,7 @@
 (function() {
   'use strict';
 
-  if (self.fetch) {
+  if (typeof window !== 'object' || window.fetch) {
     return
   }
 
@@ -108,7 +108,7 @@
   }
 
   var support = {
-    blob: 'FileReader' in self && 'Blob' in self && (function() {
+    blob: 'FileReader' in window && 'Blob' in window && (function() {
       try {
         new Blob();
         return true
@@ -116,8 +116,8 @@
         return false
       }
     })(),
-    formData: 'FormData' in self,
-    arrayBuffer: 'ArrayBuffer' in self
+    formData: 'FormData' in window,
+    arrayBuffer: 'ArrayBuffer' in window
   }
 
   function Body() {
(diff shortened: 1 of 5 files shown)


niftylettuce:
Looks like tests fail because of iframes or something?

```log
Unsafe JavaScript attempt to access frame with URL about:blank from frame with URL
```

mislav:
Sorry, but `self` was used to enable this polyfill working in [a web worker environment](https://github.com/github/fetch/blob/133bd7321019cd825989ad8dcab8c96725e0b612/test/test-worker.html).

We're sorry that you're having problems with jsdom/jsdomify, but we can't help you with that. Our polyfill works quite fine when loaded normally onto the page with `<script src>` tags. If your tools can't handle normal js files, maybe it's time to revisit your tools?

mislav:
Also see #231 #47 #31

niftylettuce:
Is there not a way you can ensure a fallback instead of `self`?  Why can't you?  This doesn't seem right.  Otherwise I have to rely on my fork to get my tests to pass for https://github.com/niftylettuce/frisbee.

niftylettuce:
`if (typeof window === 'object') self = window;`?

niftylettuce:
It really doesn't make sense to me to use `self` JUST to make your tests pass.  That's not right.

niftylettuce:
@mislav yeah I have to stop using `jsdomify` and run it per https://github.com/tmpvar/jsdom/issues/1203#issuecomment-160730614 and just use pure `jsdom` with proper `window` environment.

niftylettuce:
ignore this PR.

mislav:
> Is there not a way you can ensure a fallback instead of `self`? Why can't you? This doesn't seem right.

We welcome a PR that solves the compatibility issue you're describing in an elegant way while at the same time preserving compatibility with Web Workers. The existing CI test suite must pass.

PR #243, opened 2015-12-06
Add PATCH to normalized method names

ajwhite:
Discovered in https://github.com/facebook/react-native/issues/4594, `PATCH` is not included in the normalized method names. This lead to some problems with the lowercased `patch` being forwarded onto the OkHTTP library in React Native and throwing an error as it was not interpreted properly.

This adds consistency to the method name normalizing process by including all the supported/common method types.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -197,7 +197,7 @@
   }
 
   // HTTP methods whose capitalization should be normalized
-  var methods = ['DELETE', 'GET', 'HEAD', 'OPTIONS', 'POST', 'PUT']
+  var methods = ['DELETE', 'GET', 'HEAD', 'OPTIONS', 'POST', 'PUT', 'PATCH']
 
   function normalizeMethod(method) {
     var upcased = method.toUpperCase()


mislav:
Unfortunately, [the spec doesn't include `PATCH`](https://fetch.spec.whatwg.org/#concept-method-normalize) in the set of normalized methods.

You'll just always have to write out `PATCH` in uppercase until this is changed in the fetch spec.

ajwhite:
Interesting, alrighty then.

Lavoaster:
After reading through https://github.com/whatwg/fetch/issues/50 on this very issue, normalization is only a convenience method to help support old code. The TL;DR of it is, just write verbs in the correct way, uppercase.

I guess where some of the disconnect and confusion here is that all the examples in this are in lowercase, if they were to be presented in uppercase and the normalization hidden, then there would be less confusion.

So, would it be appropriate to update all examples so that their method names are in uppercase?

mislav:
@Lavoaster That's a good idea. Consistency in this would help people avoid the `patch` trap.

Lavoaster:
I'm wondering what @dgraham's thoughts on this are, I've just seen #272 which was rejected not too long ago.

dgraham:
Patch needs to be treated as a first-class method, with normalization, in the fetch specification to solve this problem. Shouting verbs into a readme feels like progress because we control that file, but until `patch` is treated humanely, we'll continue to spend time debugging this inconsistency in method handling.

/cc https://github.com/github/fetch/pull/272#issuecomment-177602462

Lavoaster:
The guy(s?) working on the spec for fetch, from my interpretation, doesn't see it that way. If you read through the issue I linked above they mention that normalization is, strictly speaking, breaking http standards as methods are case-sensitive and that normalization was only added for backwards compatibility - This was even noted in the spec [here](https://fetch.spec.whatwg.org/#methods) after a previous request (in 2014) to add `PATCH` to the normalization verbs was declined. The issue also talks about it not being normalized because it would be a breaking change for people relying on the case sensitivity for patch.

You hope for a future that I personally don't see forthcoming. You argue on relying on normalization which only exists for backwards compatibility, you also argue to further be non compliant with the HTTP standard which defines method names as [case sensitive](https://tools.ietf.org/html/rfc2616#section-5.1.1) by relying on a "fix" for non-compliant applications, and finally the future you're arguing for is a specification that will need to be patched every time a new verb is added. You say this is inconsistent, but this inconsistency has been caused by people not following the HTTP spec correctly in the first place, and yet you're here arguing to further that inconsistency rather than correct it.

I get that this isn't an issue for those who want to be spec compliant, but by not being spec compliant yourself, you're pushing people into this trap because your examples are being used as a reference for the rest of the internet on how to use fetch, as this package is what allows its use today. Examples in the fetch [spec](https://fetch.spec.whatwg.org/#fetch-api) use uppercase method names, and for that matter all references to method names are in uppercase in the spec, The [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Supplying_request_options) on fetch uses uppercase method names.

This won't solve entirely solve the problem of normalization, but it will improve the situation.

Lavoaster:
@dgraham I'm not prepared to let this issue just fade away. If you're insistent on not changing the examples here and hiding normalization, then why haven't you logged an issue to change the spec?

jmeas:
@dgraham , do you agree that the current README is setting developers up to waste time debugging this issue? If so, there are steps you can take today to help alleviate or even eliminate that problem (the one of wasted developer time) altogether.

Keeping the verbs lowercase in the README without mention of how they actually should be (and in some cases *need* to be uppercase) is _creating_ this new problem. And it's doing nothing to help fix the spec that you believe should be changed.

I think there are ways to update the README that communicates your opinions while also being pragmatic. Consider:

> #### A Note on Verbs
>
> The examples in the README use lowercase HTTP verbs. Although it'd be nice if we
> could use lowercase all of the time, the spec writers disagree with me on that point. For
> other verbs, like PATCH, you'll need to be sure to use uppercase letters.
> 

Although it'd be nice, you can't will the spec to change by rebelling against uppercase verbs in the README. The only consequence of this action is wasting the :clock1130: of lots of people.

mislav:
Current README examples do give the reader the impression that HTTP method can be supplied in lowercase. I can totally understand how somebody would be surprised that `patch` isn't included in this normalization. I'm 👍 on accepting a PR that uppercases our examples, unless @dgraham thinks that effort would be futile.

But please, don't get too worked up on this issue. Even though a subset of developers might get bitten at some point, the situation isn't that dire.

Lavoaster:
FYI - There is now an issue to add console warnings whenever normalization is invoked. https://github.com/whatwg/fetch/issues/259

Issue #214, opened 2015-10-05
Should not restrict the status code


micky2be:
At the following lines it's clear that the status code is enforced to be between 100 and 599.
https://github.com/github/fetch/blob/master/fetch.js#L313-L316

But in some cases servers/APIs can return a custom status code higher than 599.

mislav:
I'm not sure if we should officially support status codes higher than 599, since those don't seem to be part of the HTTP spec. [HTTP/1.1 only defines](https://pretty-rfc.herokuapp.com/RFC2616#status.code.and.reason.phrase) the first digit to be in the range of 1–5.

micky2be:
That should be up to the developers using the fetch library. 
We already have to check for 2xx anyway. 
Getting blocked beforehand doesn't do any good for the people expecting receiving this status.

mislav:
We do this check mostly because of IE, which returns large error codes to signal various network errors. If you want to support status codes > 599, you have to come up with an idea of how to detect IE network errors in another way.

micky2be:
Why would fetch need to filter it??
Like I said, let the developers using fetch do it. 
Fecth on Chrome doesn't filter. Node doesn't filter. So why this library do?
Why adding limitation?
Doing the following by a user/developer should be enough to me. 


```js
function checkStatus(response) {
  if (response.status >= 200 && response.status < 300) {
    return response
  } else {
    var error = new Error(response.statusText)
    error.response = response
    throw error
  }
}
fetch(api).then(checkStatus);
```

ronkorving:
This is a polyfill, so I would expect it to stick to standards (the standards for the fetch API, not HTTP), or if those are lacking in clarity, to do what the majority of browsers do. If that is to let the 6xx go through, I think this library should too, right? Limiting developers from receiving these codes can be a blocker. The opposite case cannot be made.

mislav:
> Why would fetch need to filter it??

@micky2be Please read my above comment again:

> We do this check mostly because of IE, which returns large error codes to signal various network errors. If you want to support status codes > 599, you have to come up with an idea of how to detect IE network errors in another way.

So if we find a reliable way to detect IE network errors, we can remove the filter.

dgraham:
Good discussion about two competing concerns: passing *intentionally* invalid status codes from the server through to the client vs supporting IE 9's invalid client status codes. I'm closing this out though, because supporting IE 9 is more important right now.

[...]

PR #247, opened 2015-12-11
Accept HTTP codes over 599

stelcheck:
Ref: https://github.com/github/fetch/issues/214

I understand this would not be mergeable until later, but thought I would make a PR for it anyway, as a note.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -342,7 +342,7 @@
 
       xhr.onload = function() {
         var status = (xhr.status === 1223) ? 204 : xhr.status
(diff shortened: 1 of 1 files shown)


mislav:
We'd rather not keep around open PRs that are not mergeable in their current state.

Since we're still supporting IE 9 and this PR doesn't address the concerns outlined in https://github.com/github/fetch/issues/214#issuecomment-146509126, I'm gonna go ahead and close this. This and related changes will be welcome after dropping IE 9 support, but right now we don't have a timeframe for that.

Issue #250, opened 2015-12-19
would you guys support for typescript?


Brooooooklyn:
I have used this library in Typescipt, and I'm wondering if you have any plan to have a .d.ts file to support Typescript?

```
declare module 'whatwg-fetch' {
  namespace fetch {}
  function fetch(input: any, init?: Fetch.IRequestOptions): Fetch.Promise<Fetch.ResponseBody>
  export = fetch
}

declare module Fetch {
  interface IInput {
    bodyUsed: boolean
    url: string
    credentials: string
    method: string
    mode: string
    _bodyInit: boolean
  }
  interface IRequestOptions {
    body?: string
    headers?: Headers
    credentials?: string
    mode?: string
  }

  interface IResponseOptions {
    type?: string
    status?: number
    statusText?: string
    headers?: Headers
    url?: string
  }

  class Headers {
    constructor(header: any)
    append(name: string, value: any): void
    delete(name: string): void
    get(name: string): any
    getAll(name: string): any[]
    has(name): boolean
    set(name: string): void
    forEach(callback: (thisArg: any, value: any, name: string, context: any) => any, thisArg: any): void
  }

  class Request {
    constructor(input: IInput, options?: IRequestOptions)
    clone(): Request
  }

  class Response {
    constructor(bodyInit: any, options?: IResponseOptions)
    clone(): Response
    static error(): Request
    static redirect(): Request
  }

  class ResponseBody {
    bodyUsed: boolean
    _initBody(body: any): void
    _bodyText: string
    _bodyBlob: Blob
    _bodyFormData: FormData
    blob(): Promise<Blob>
    arrayBuffer(): Promise<Blob>
    text(): Promise<string>
    formData(): Promise<FormData>
    json(): Promise<any>
  }

  interface IteratorResult<T> {
    done: boolean
    value?: T
  }

  interface Symbol {
    toString(): string

    valueOf(): Object

    [Symbol.toStringTag]: string
  }

  interface SymbolConstructor {
    prototype: Symbol

    (description?: string|number): symbol

    for(key: string): symbol

    keyFor(sym: symbol): string

    hasInstance: symbol

    isConcatSpreadable: symbol

    iterator: symbol

    match: symbol

    replace: symbol

    search: symbol

    species: symbol

    split: symbol

    toPrimitive: symbol

    toStringTag: symbol

    unscopables: symbol
  }

  var Symbol: SymbolConstructor

  interface Iterator<T> {
    next(value?: any): IteratorResult<T>
[...]

PR #251, opened 2015-12-22
feat(d.ts): add type definition

Brooooooklyn:
https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages

typings support for Typescript.

And you should publish the fetch.d.ts to npm.

### DIFF
--- a/fetch.d.ts
+++ b/fetch.d.ts
@@ -1,0 +1,160 @@
+interface IInput {
+  bodyUsed: boolean
+  url: string
+  credentials: string
+  method: string
+  mode: string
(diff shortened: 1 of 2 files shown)


mislav:
Thanks, but this is too much for us to maintain over time, as we don't have any familiarity nor use for TypeScript. This is best maintained in an external repository. You could, for instance, publish your own npm module called `whatwg-fetch-typescript` that would be a fork of this repo with this extra file.

Closing per @dgraham's comment in https://github.com/github/fetch/issues/250#issuecomment-166005679

PR #253, opened 2016-01-06
Handle cases where `self` isn't defined

### LINKED ISSUE #252: Possibility for a 'self' polyfill
In non-browser execution environments, `self` typically doesn't exist. It seems to me that to avoid errors while loading the script, it'd be nice if fetch would failover to use an equivalent global instead - only if `self` isn't available.

An additional benefit is that this would improve support for very old browsers which don't have `self` (which was added in HTML5).

The purpose of this change is not to provide support for non-browser environments, but to avoid crashing those environments when the script is included. This happens, for example, when including `isomorphic-fetch` in React Native due to the way it packages the node modules.

Would you be open to considering a PR which implements this?

billinghamj:
As per: https://github.com/github/fetch/issues/125#issuecomment-169118755

Resolves #252, #238, #165, #125, #124, and probably a few others.

There are a lot of different ways this code could be written. If you have style preferences contrary to the way I've done this, I'd be very happy to change it.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,4 +1,4 @@
-(function() {
+(function(self) {
   'use strict';
 
   if (self.fetch) {
@@ -378,4 +378,4 @@
     })
   }
   self.fetch.polyfill = true
-})();
+})(typeof self !== 'undefined' ? self : this);


mislav:
Looks good!

@dgraham Any objections?

dgraham:
Invoking the wrapper function with `this` resolves to `window`, which is how we started all the way back in c1af6de. This looks good to me. :shipit:

chrisirhc:
Wouldn't this still throw a `ReferenceError` exception where `self` isn't defined? How about doing a typeof check like `typeof self !== 'undefined' ? self : this` ?

mislav:
Good point @chrisirhc. Someone should test this out in an isomorphic environment and report back. I have no idea how it would handle the `self || this` construct.

billinghamj:
Hmm, yes that is true. Another option could be:

```js
this.self || this
```

Which works exactly the same way but should not throw.

billinghamj:
I have updated the code as per the above comment.

mislav:
Can someone test this out in an isomorphic environment, as per people's complaints on aforementioned threads?

billinghamj:
Node - master branch:

```
$ node
> var a = require('./fetch')
ReferenceError: self is not defined
    at /Users/james/Projects/fetch/fetch.js:4:7
    at Object.<anonymous> (/Users/james/Projects/fetch/fetch.js:381:3)
    at Module._compile (module.js:398:26)
    at Object.Module._extensions..js (module.js:405:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at repl:1:9
    at REPLServer.defaultEval (repl.js:254:27)
```

Node - this PR:

```
$ node
> var a = require('./fetch')
undefined
> a
{ Headers: [Function: Headers],
  Request: [Function: Request],
  Response: { [Function: Response] error: [Function], redirect: [Function] },
  fetch: { [Function] polyfill: true } }
```

React Native - [master](http://imgur.com/Twm6QGK)
React Native - [this PR](http://imgur.com/1terliO) (this new error is a separate issue from [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch/blob/master/fetch-npm-browserify.js#L6))

I can't speak for other environments though

fetch.js
```diff
@@ -378,4 +378,4 @@
     })
   }
   self.fetch.polyfill = true
-})();
+})(this.self || this);
```
mislav: I feel like I would like to revert to @chrisirhc's idea by doing

```js
typeof self !== 'undefined' ? self : this
```

simply because that approach has been better battle-tested by compilers such as Browserify.

billinghamj:
@mislav updated as-per your comment

mislav:
Thank you!

yorkie:
Hey @mislav did you forget to publish to npm, when I install 0.11.0 get the following:

```
pm ERR! Darwin 15.2.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "i"
npm ERR! node v5.0.0
npm ERR! npm  v3.3.6

npm ERR! No compatible version found: whatwg-fetch@0.11.0
npm ERR! Valid install targets:
npm ERR! ["0.10.1","0.10.0","0.9.0","0.8.2","0.8.1","0.8.0","0.7.0","0.6.1","0.6.0","0.5.0"]
npm ERR!
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/yorkieliu/workspace/weflex/gian/npm-debug.log
```

mislav:
@yorkie Something's wrong with our build system and auto-publish on npm. Stay tuned.

yorkie:
I see, is there something I can help? This is no hurry, but if we can get a working notification with npm, that would be so sweet :-)

mislav:
I just did a manual `npm publish` for now.

yorkie:
It's working now, very thanks to you @mislav :+1:

PR #255, opened 2016-01-08
Make fetch add a `Content-Type` header based on the type of the body.

xymostech:
Summary: The process of extracting a body from a passed in body object
(described here: https://fetch.spec.whatwg.org/#concept-bodyinit-extract)
creates both a byte stream and a corresponding `Content-Type` value. These two
values are used in the constructor of `Request` and `Response` objects to
initialize the body and to set an initial `Content-Type` header.

This change makes `Body._initBody()` return the `Content-Type` value
corresponding to the type of the object passed in, and to set that header in
the corresponding `Request` and `Response` objects that are being constructed.

There's a little bit of a dance I need to do in Request to keep track of where
the body is coming from, because we're only supposed to be setting the
`Content-Type` header if we're getting the body from the options passed in, not
if we're copying it from another `Request`. (This was caught by one of the
tests. Yay tests!)

I added some corresponding tests to make sure this works.

Test Plan:
I'm a little bit confused about how to run the tests normally, but this is what
I did:

 - Run `make lint`, see that everything passes
 - Start up the test server by running `node script/server 3900 &`
 - Visit http://localhost:3900/test/test.html and
   http://localhost:3900/test/test-worker.html in Chrome and see that all of
   the tests pass.

From what I can tell, running the tests in Chrome just used the native fetch
objects, so to test the polyfill I added `window.fetch = null` in
test/test.html before it includes '../fetch.js', and added `self.fetch = null`
in test/worker.js before it imports '../fetch.js'. Then:

 - Visit http://localhost:3900/test/test.html and
   http://localhost:3900/test/test-worker.html in Chrome and see that all of
   the tests pass.

Reviewers: @jeresig

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -128,10 +128,16 @@
       this._bodyInit = body
       if (typeof body === 'string') {
         this._bodyText = body
+        return 'text/plain;charset=UTF-8'
       } else if (support.blob && Blob.prototype.isPrototypeOf(body)) {
         this._bodyBlob = body
+        if (body.type !== '') {
+          return body.type
+        }
       } else if (support.formData && FormData.prototype.isPrototypeOf(body)) {
         this._bodyFormData = body
+        // Since we don't know what the multipart/form-data boundary string
+        // will be, we can't set a Content-Type here
       } else if (!body) {
         this._bodyText = ''
       } else if (support.arrayBuffer && ArrayBuffer.prototype.isPrototypeOf(body)) {
@@ -207,6 +213,7 @@
   function Request(input, options) {
     options = options || {}
     var body = options.body
+    var bodyFromOptions = true
     if (Request.prototype.isPrototypeOf(input)) {
       if (input.bodyUsed) {
         throw new TypeError('Already read')
@@ -221,6 +228,7 @@
       if (!body) {
         body = input._bodyInit
         input.bodyUsed = true
+        bodyFromOptions = false
       }
     } else {
       this.url = input
@@ -237,7 +245,10 @@
     if ((this.method === 'GET' || this.method === 'HEAD') && body) {
       throw new TypeError('Body not allowed for GET or HEAD requests')
     }
-    this._initBody(body)
+    var contentType = this._initBody(body)
+    if (bodyFromOptions && contentType && !this.headers.get('Content-Type')) {
+      this.headers.set('Content-Type', contentType)
+    }
   }
 
   Request.prototype.clone = function() {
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for the PR. Why did you close it?

Is this per whatwg fetch specification? Do native browser implementations behave in a similar way when setting implicit `content-type` values?

> From what I can tell, running the tests in Chrome just used the native fetch

Yeah, that's the problem with the test suite right now: if the browser already implements `window.fetch`, the polyfill code won't be exercised. Your workaround is fine for now.

xymostech:
Hello! I closed it mostly because I intended it as a pull request to our fork of the polyfill, but it sent it here by default! I can reopen it if you think it's useful!

Yes, this is per whatwg fetch specification (see https://fetch.spec.whatwg.org/#dom-request step 32, and https://fetch.spec.whatwg.org/#dom-response step 7 which set the `content-type` header, which refer to https://fetch.spec.whatwg.org/#concept-bodyinit-extract for finding the correct content type). Chrome at least does behave in a similar way; I can (and probably should) test in other browsers to see if they do the same (looks like firefox is the only other browser that implements it?).

xymostech:
Hmm, not sure why that test is failing, looks like it's timing out.

mislav:
At a glance this looks :+1:, but I'll need to do a deeper review. Thanks!

mislav:
Thanks for your contribution! I've made some tweaks and managed to distill the implementation down to just:

```diff
diff --git a/fetch.js b/fetch.js
index f89f27a..3792adf 100644
--- a/fetch.js
+++ b/fetch.js
@@ -140,6 +140,14 @@
       } else {
         throw new Error('unsupported BodyInit type')
       }
+
+      if (!this.headers.get('content-type')) {
+        if (typeof body === 'string') {
+          this.headers.set('content-type', 'text/plain;charset=UTF-8')
+        } else if (this._bodyBlob && this._bodyBlob.type) {
+          this.headers.set('content-type', this._bodyBlob.type)
+        }
+      }
     }
 
     if (support.blob) {
@@ -276,13 +284,13 @@
       options = {}
     }
 
-    this._initBody(bodyInit)
     this.type = 'default'
     this.status = options.status
     this.ok = this.status >= 200 && this.status < 300
     this.statusText = options.statusText
     this.headers = options.headers instanceof Headers ? options.headers : new Headers(options.headers)
     this.url = options.url || ''
+    this._initBody(bodyInit)
   }
 
   Body.call(Response.prototype)
```

xymostech:
Looks great to me! Thanks!

PR #258, opened 2016-01-10
Exercise both polyfill and native `fetch` in test suite

mislav:
Before, in browsers that had native `fetch` implementation the polyfill code was never exercised by the test suite. This ensures that both implementations are exercised.

Bonus features: better first-run experience when cloning the project and trying to run `script/server`:

* The port number is now printed on stderr;
* Test suite now renders at root path `/` instead of having to type out `/test/test.html`;
* Helpful error message is displayed if you forgot to run `make`.


dgraham:
🚢

mislav:
This is not yet ready. Locally I've upgraded mocha to get access to `skip()` test or suite at runtime, as opposed to definition time, but the feature doesn't work as advertised. I'll need more time.

mislav:
@dgraham This is now ready for another :eyes:, thanks. I've also brought in the Firefox fix from #157 because JSON parsing exception somehow caused cascading failure in Mocha test output, ruining the rest of HTML rendering and causing failures in unrelated tests.

dgraham:
> I've also brought in the Firefox fix from #157 because JSON parsing exception somehow caused cascading failure in Mocha test output

In general, I'm not a fan of patching production code to accommodate the test suite. Is there another way the tests can avoid this problem?

package.json
```diff
@@ -13,10 +13,10 @@
   "devDependencies": {
     "bower": "1.3.8",
     "chai": "1.10.0",
-    "jshint": "2.5.2",
+    "jshint": "2.8.0",
```
staabm: Shouldnt you define min-versions instead of fixed ones, e.g. Using caret operator `^`?

mislav:
> Is there another way the tests can avoid this problem?

We could skip the test in the case we're testing with Firefox and we're exercising the polyfill code. Is that what you would want?

package.json
mislav: Usually, but in this case I don't want jshint dependency to move around at all, lest its behavior suddenly changes, causing failures in CI for unsuspecting contributors and their pull requests.

dgraham:
Skipping this test in Firefox sounds good to me. We're testing the native implementations only to ensure our polyfill behavior matches. We're covered in that case with Chrome.

PR #265, opened 2016-01-22
Check forEach method of Headers that's not implemented in some browsers

yborunov:
In Firefox v43.0.3 there's no forEach method for Headers interface which causes errors in libraries that rely on this method when using fetch API.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,7 +1,16 @@
 (function(self) {
   'use strict';
 
-  if (self.fetch) {
+  // In some browsers fetch API is partially implemented which stops it from polyfilling
+  //   while keeps implementation that's lack of some essential methods,
+  //     such as forEach method of Headers interface at the moment which causes problems in Firefox browser v43.0.3 (01-22-2016)
+  // Current specification of Headers interface: https://developer.mozilla.org/en-US/docs/Web/API/Headers
+  function allMethodsAvailable() {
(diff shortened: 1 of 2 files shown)


yborunov:
Here's an example of library Restful.JS that doesn't work in Firefox because of that: https://github.com/marmelab/restful.js/issues/66

dgraham:
It's best to avoid methods that all implementations have not yet added. This polyfill is written to the Fetch specification, which includes a `forEach` method.

yborunov:
If polyfill provides `forEach` method, what the point of suggesting to avoid using it?
`Restful.js` library was mislead by this polyfill which has this implementation.

dgraham:
`Headers#forEach` isn't really usable until it's [added to Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1216751).

mislav:
Ideally, this polyfill should detect that `Headers#forEach` method is missing and polyfill it in. I would want to do that, but we can't add `forEach` to Headers since we have no means to iterate over all headers stored in the object.

antalutrazvan:
the issue persists in Internet Explorer all versions, because Headers.keys is not implemented.
Please fix this, by polyfilling it.

yborunov:
@antalutrazvan what version of IE are you using? Versions <= 11 should be polyfilled easily since IE <=11 doesn't have Fetch API implementation.

PR #270, opened 2016-01-30
Drop IE-specific status codes workarounds and require IE10+

### LINKED ISSUE #214: Should not restrict the status code

mislav:
Fixes #214


dgraham:
:shipit:

mislav:
Current version is 0.11. Are we ready to release 1.0.0?

dgraham:
> Are we ready to release 1.0.0?

Yes, fetch has been stable in production for over a year.

Issue #254, opened 2016-01-07
http method patch is case sensitive


kocur4d:
Is this intended?

https://github.com/github/fetch/blob/master/fetch.js#L200

method patch is not the member of the array.

spend some time till I figure out why my fetch was throwing:

```
Fetch API cannot load http://localhost:3000/api/v1/tasks. Method patch is not allowed by Access-Control-Allow-Methods in preflight response.
```

dgraham:
Unfortunately, it is intended. `PATCH` must be shouted in all caps. See https://github.com/github/fetch/issues/37 for more discussion.

kocur4d:
ah I don't know how I missed this issue...

Anyway it might be worth to add a small paragraph to the readme I literally spend an hour on this - but on the other hand I missed the issue #37 and didn't check source earlier...

noise-machines:
What do do you think of logging a warning when fetch is called with lowercase http verbs? Just dropped over an hour on what seemed like a very mysterious bug.

Issue #37, opened 2014-11-21
Add PATCH HTTP verb normalization


h2non:
Could you add the PATCH HTTP verb? It is supported from XHR2

PR #272, opened 2016-01-31
Use uppercase methods in README

jmeas:
HTTP methods are case-sensitive. The examples in the README currently
lead developers to think that this API normalizes all of the method
names to uppercase, leading to issues like #37 and #254.

Setting the right precedent in the README will prevent developers
from making mistake of trying to use method: 'patch' (or any
other verb that isn't normalized for backwards compatibility).

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ fetch('/users.json').then(function(response) {
 var form = document.querySelector('form')
 
 fetch('/users', {
-  method: 'post',
+  method: 'POST',
   body: new FormData(form)
 })
 ```
@@ -86,7 +86,7 @@ fetch('/users', {
 
 ```javascript
 fetch('/users', {
-  method: 'post',
+  method: 'POST',
   headers: {
     'Accept': 'application/json',
     'Content-Type': 'application/json'
@@ -108,7 +108,7 @@ data.append('file', input.files[0])
 data.append('user', 'hubot')
 
 fetch('/avatars', {
-  method: 'post',
+  method: 'POST',
   body: data
 })
 ```


jmeas:
I tried to find the GitHub CLA on the interwebs, but couldn't find it. If someone sends me a link I can get that signed.

dashaus:
I think that is here: https://cla.github.com

jmeas:
Thanks @dashaus ! Now I'm really embarrassed. I can't figure out how to sign the thing on the page. I thought maybe there was a JavaScript timer on the page hiding the accept button for a minute, but it's not sending over any JS.

I'm wondering if this empty div is where the sign button is supposed to show up?

image

jmeas:
Oh, figured it out! I needed to click through to the CLA from the failed PR status. Kinda weird, but hey, I figured it out :v:

dgraham:
The case of `PATCH` is indeed unfortunate. Perhaps one day it will be afforded the privileges enjoyed by the original Gang of Four verbs and may be spoken in an inside voice. For now, we must shout it into the computer and hope it reaches the server.

I continue to hope for a `patch` future.

jmeas:
@dgraham I, too, share a vision with you for patch! But alas, it may be some time before we get there. In the meantime, I think we have a responsibility to do what we can to prevent developers from spending time figuring out why (seemingly arbitrary) HTTP verbs don't work with this library.

I think it's a problem that the README sets developers up to fail right now. If they're like me, they probably copy + paste an example, then modify it for their own needs. Unfortunately, nothing in the README explains the unfortunate detail that only *some* of the verbs are normalized.

It's clear from those two issues (#37 and #254) – and my own experience – that some updated documentation could do some good here. Would you be open to a PR that adds a note about case sensitivity to the README?

ajwhite:
Thanks @dgraham :+1:

jmeas:
@dgraham , thanks, yo :v:

Lavoaster:
:tada:

Thanks @dgraham :heart:

PR #278, opened 2016-02-11
Use response.ok in readme

jacobrask:
Use the `ok` property instead of explicitly checking for status code. A lot of people seem to copy paste this example, which means that they miss out on the useful `ok` property.

See https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ status, define a custom response handler:
 
 ```javascript
 function checkStatus(response) {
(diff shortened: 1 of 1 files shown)


dgraham:
The `ok` property was [added after](https://github.com/github/fetch/pull/82) `fetch` first shipped in Chrome. So it's best to use the status code checks until those old versions of Chrome fall out of usage.

PR #282, opened 2016-02-18
fix example

vga-:
fixes example to use api.github.com rather than github.com

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -6,7 +6,7 @@
 </head>
 <body>
   <script>
-    var result = fetch('https://github.com')
+    var result = fetch('https://api.github.com')
 
     result.then(function(response) {
       console.log('response', response)


mislav:
The example URL is just for demonstration purposes only. It doesn't really matter what the exact URL is. People are to substitute their own URL in this field. The URL `github.com` works well as an example because it's simple and also refers to the entity behind this project.

vga-:
Shouldn't the example at least work if someone copies and pastes it?

mislav:
You're right! `api.github.com` allows for CORS and therefore will work when `github.com` doesn't. Thanks!

mislav:
@dgraham Ah, thanks for that bit of context. @vga- It would have been helpful to me if you linked to the prior discussion when opening the PR. More context around the proposed change always helps!

vga-:
Okay, thanks! I'll make sure to do that in the future.

PR #283, opened 2016-02-24
Added exports to the polyfill fetch

pinturic:
This way it is possible to use the polyfill even if self.fetch is
already present. There are cases (test environment, compatibility tests,
etc..) in which you want to use the polyfill even if you already have a native fetch.


dgraham:
Thanks, but this isn't something we're interested in supporting.

Issue #275, opened 2016-02-09
Usage with babel and ES6 


abimelex:
I try to use the package with babel and es6
`import fetch from 'whatwg-fetch';`

Unfortunately I get the following error:

`Uncaught TypeError: (0 , _whatwgFetch2.default) is not a function`

TrySound:
@abimelex 
Description 
> A window.fetch JavaScript polyfill.
fetch do not export modules.
You can do
```js
import 'whatwg-fetch';
// And then just use global variable.
fetch(...);
```

abimelex:
Ah thank you, thats perfect.

TrySound:
Don't forget to close issue if you found answers.

webjay:
I solved it with `import 'isomorphic-fetch';` in my `jest-setup.js`.

PR #287, opened 2016-03-05
Add ES2015+ example

jamonholmgren:
I had the same question as @abimelex in #275, and figured it would be useful to have a short example in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,13 @@ For a node.js implementation, try [node-fetch](https://github.com/bitinn/node-fe
 
 For use with webpack, refer to [Using WebPack with shims and polyfills](http://mts.io/2015/04/08/webpack-shims-polyfills/).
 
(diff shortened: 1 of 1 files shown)


mislav:
I don't think that the instructions for loading this polyfill are different from any other polyfill out there (no exports). Therefore I don't find this particularly useful. But if people are constantly looking this up, then I guess it can't hurt to clear this up. Thoughts @dgraham?

jamonholmgren:
You're right, it's no different than other polyfills...this was the first polyfill I've used within our new Babel environment, so it was the first time I'd considered this question.

dgraham:
Several people have found the answer in #275 useful, so promoting it to the README might help people importing modules for side effects, like adding window global polyfills.

jamonholmgren:
Thank you!

PR #289, opened 2016-03-08
Guard against `xhr.getAllResponseHeaders()` being `null`

steveluscher:
From https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#getAllResponseHeaders():

> Returns all the response headers, separated by CRLF, as a string, or null if no response has been received.

I discovered this on Android, once when `xhr.getAllResponseHeaders()` evaluated to `null`.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -267,7 +267,7 @@
 
   function headers(xhr) {
     var head = new Headers()
-    var pairs = xhr.getAllResponseHeaders().trim().split('\n')
+    var pairs = (xhr.getAllResponseHeaders() || '').trim().split('\n')
     pairs.forEach(function(header) {
       var split = header.trim().split(':')
       var key = split.shift().trim()


fetch.js
```diff
@@ -267,13 +267,16 @@
 
   function headers(xhr) {
     var head = new Headers()
-    var pairs = xhr.getAllResponseHeaders().trim().split('\n')
```
dgraham:
```js
var pairs = (xhr.getAllResponseHeaders() || '').trim().split('\n')
```

mislav: I agree that this approach would be more elegant than wrapping everything in a `typeof` check.

steveluscher: Sure. I was just trying to avoid creating and `forEach`-ing that array for nothing.

dgraham:
@steveluscher This looks great. Can you squash this into a single commit? If you accept the contributor license agreement by clicking the `Details` link next to the failed build status, I can merge this branch.

steveluscher:
cc/ @jamesgpearce

dgraham:
Thanks, @steveluscher!

PR #295, opened 2016-03-12
Add Headers iterators

### LINKED ISSUE #292: Getting all headers, mismatch between the specs and polyfill and implementations
So MDN says to use
```
headers.entries();
```
the fetch spec says the Headers object is iterable itself (so it should have a `Symbol.iterator` function)

and then the polyfill has `forEach`

Looking at chromes fetch we have

image

and firefox has

image

so I guess to be forwards compatible I have to try to use forEach and fall-back to either an iterator on headers itself or entries..

Should an iterator be added and forEach removed so that people donm't think that forEach is guarenteed to be present when not polyfilled?

[Edited to be more correct about iterator on headers directly]

dgraham:
Adds `entries`, `keys`, and `values` iterator functions to `Headers`. Each iterator, and `Headers` itself, is also `iterable` if the browser supports `Symbol.iterator`. The [medikoo/es6-symbol](https://github.com/medikoo/es6-symbol) polyfill may be used in browsers without `Symbol`.

Background:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#iterable
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators
- https://fetch.spec.whatwg.org/#headers
- http://heycam.github.io/webidl/#idl-iterable
- http://heycam.github.io/webidl/#es-iterable

Closes #292.

/cc @lukeapage @dkfiresky

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -5,6 +5,20 @@
     return
   }
 
+  var support = {
+    iterable: 'Symbol' in self && 'iterator' in Symbol,
+    blob: 'FileReader' in self && 'Blob' in self && (function() {
+      try {
+        new Blob()
+        return true
+      } catch(e) {
+        return false
+      }
+    })(),
+    formData: 'FormData' in self,
+    arrayBuffer: 'ArrayBuffer' in self
+  }
+
   function normalizeName(name) {
     if (typeof name !== 'string') {
       name = String(name)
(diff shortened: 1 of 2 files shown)


lukeapage:
Looks good to me, thanks for adding this. Do you think `entries`, `values`, `keys` will be added to the spec at some point?

dgraham:
> Do you think entries, values, keys will be added to the spec at some point?

It's a little confusing, but the iterator functions are actually specified by fetch. The `Headers` class implements the `iterable` interface, which is defined in a separate specification. All specs are linked in the initial comment above.

lukeapage:
Ah i see, thanks.

PR #299, opened 2016-03-25
README examples with uppercased method names

ajwhite:
This addresses the confusion brought up in #243. 

TL;DR - [method types are normalized to be uppercased](https://github.com/ajwhite/fetch/blob/c5200ebe021d1ed0222d6005641ecac35e29e589/fetch.js#L199-L205), however [`PATCH` is not included in the normalization rule of the `fetch` specification](https://fetch.spec.whatwg.org/#concept-method-normalize). To avoid confusion, the examples now show method names in uppercase.

Hopefully this can help train the developer to use uppercase names and not fall into the `patch` trap.

/cc @mislav and @dgraham

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ fetch('/users.json').then(function(response) {
 var form = document.querySelector('form')
 
 fetch('/users', {
-  method: 'post',
+  method: 'POST',
   body: new FormData(form)
 })
 ```
@@ -86,7 +86,7 @@ fetch('/users', {
 
 ```javascript
 fetch('/users', {
-  method: 'post',
+  method: 'POST',
   headers: {
     'Accept': 'application/json',
     'Content-Type': 'application/json'
@@ -108,7 +108,7 @@ data.append('file', input.files[0])
 data.append('user', 'hubot')
 
 fetch('/avatars', {
-  method: 'post',
+  method: 'POST',
   body: data
 })
 ```


jmeas:
Related: #272

mislav:
Huh. I didn't realize that there was already a PR exactly like this. If we decide this was a good idea, I'd rather merge the first PR to give the original author proper credit.

ajwhite:
Ah - nor did I. I share the same feelings @mislav.

Can we revisit #272?

Lavoaster:
Has there been any internal discussion about this?

mislav:
Yes, but @dgraham and I are in disagreement. He prefers to keep method names lowercase in our examples for aesthetic reasons. We understand that this might be tripping people out when they use `patch`, but until we reach some kind of consensus, I'm afraid this will have to be put on hold for now.  Thanks for the suggestion, though!

jmeas:
@mislav, in another issue, you posted that this isn't the biggest deal in the world. I agree with you. At the same time, on principle it astounds me that someone would place aesthetic concerns over practicality when it comes to a README – especially when there's evidence that the README is a source of confusion.

So far, the solutions PR'ed have focused on the most useful, but also the ugliest (I guess) change, which would be capitalizing the verbs everywhere. But there are in-betweens that can preserve @dgraham's preference while also helping folks out. I posted this in another issue, but what about something like:

> #### A Note on Verbs

> The examples in the README use lowercase HTTP verbs. Although they look nicer when
> lowercase, the spec only allows for certain verbs to be lowercase. For most verbs, like
> PATCH, you'll need to make sure to use uppercase letters.

If I PR'd that could we get it merged?

PR #304, opened 2016-03-30
Add support for URLSearchParams POST body

### LINKED ISSUE #296: It will be good to add URLSearchParams support
To make it possible to use x-www-urlencoded data.

Thanks.

dgraham:
Allows a [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) object to be serialized as a request body with a `Content-Type: application/x-www-form-urlencoded`  header.

```js
var params = new URLSearchParams()
params.append('user', 'hubot')
fetch('/test', {method: 'post', body: params})
```

Related:

- https://fetch.spec.whatwg.org/#body-mixin
- https://url.spec.whatwg.org/#urlsearchparams
- https://developers.google.com/web/updates/2016/01/urlsearchparams
- https://github.com/WebReflection/url-search-params

Closes #296.

/cc @mislav @btd

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -6,6 +6,7 @@
   }
 
   var support = {
+    searchParams: 'URLSearchParams' in self,
     iterable: 'Symbol' in self && 'iterator' in Symbol,
     blob: 'FileReader' in self && 'Blob' in self && (function() {
       try {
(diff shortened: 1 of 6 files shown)


mislav:
Looks good!

fetch.js
```diff
@@ -193,6 +194,8 @@
         this._bodyBlob = body
       } else if (support.formData && FormData.prototype.isPrototypeOf(body)) {
         this._bodyFormData = body
+      } else if (support.searchParams && URLSearchParams.prototype.isPrototypeOf(body)) {
```
Mouvedia: Would this still work if the `body` came from another realm?

mislav: Maybe not. To be sure, one could perform search params serialization manually via `toString()` when passing it to `fetch()`.

PR #306, opened 2016-04-01
Reject promise on request timeout

### LINKED ISSUE #294: Neither then() nor catch() executed when request timeout in Safari
In Safari V9.0.3 (10601.4.4), neither `then()` nor `catch()` is called when a request timeout occurs. Instead, Safari prints out the following error to the console:

screen shot 2016-03-11 at 3 43 22 pm

In Chrome V49.0.2623.87 (64-bit), the `catch()` callback is properly executes as expected. 

Test Script Used:

```
fetch('http://10.255.255.1') // a non-routable IP address
    .then(function(response) {
        console.log("response.ok", response.ok);
    }).catch(function(error) {
        console.log('Failed: ', error)
    });
```

Without either `then()` or `catch()` being called on timeout, how are we supposed to handle this type of request failure for Safari browsers?

dgraham:
Firefox and Chrome trigger the request's `onerror` handler when a timeout occurs. Safari triggers the `ontimeout` handler.

This can be observed by making a request to an unroutable address:

```js
xhr.onload = console.log.bind(console, 'loaded')
xhr.onerror = console.log.bind(console, 'errored')
xhr.ontimeout = console.log.bind(console, 'timeout')
xhr.open('GET', 'http://10.255.255.1')
xhr.send()
```

Fixes #294.

@mislav @fabioberger

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -429,6 +429,10 @@
         reject(new TypeError('Network request failed'))
       }
 
+      xhr.ontimeout = function() {
+        reject(new TypeError('Network request failed'))
+      }
+
       xhr.open(request.method, request.url, true)
 
       if (request.credentials === 'include') {


fabioberger:
Thanks for putting this together dgraham! I hope it gets merged soon!

staabm:
Would be great to have it documented in the source which lines covering which browser

mislav:
👍 🚢

PR #314, opened 2016-04-14
Switch to `mocha-phantomjs-core` and system PhantomJS

mislav:
This switches to the more lightweight mocha-phantomjs-core project that allows us to use system PhantomJS, which should be the later 2.x series rather than 1.9.

/cc @dgraham

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
     "chai": "1.10.0",
     "jshint": "2.8.0",
     "mocha": "2.1.0",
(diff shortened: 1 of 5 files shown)


mislav:
Hmm [seems like](https://travis-ci.org/github/fetch/jobs/122980282) Travis CI still has PhantomJS on 1.9.8. But at least locally you're free to test on whatever is the system version. Passes for me on 2.1.1

dgraham:
:shipit:

PR #316, opened 2016-04-15
README: + matthew-andrews/isomorphic-fetch

tomByrer:
[No blob](https://github.com/bitinn/node-fetch/issues/78) support in node-fetch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ This can also be installed with `npm`.
 $ npm install whatwg-fetch --save
 ```
 
-For a node.js implementation, try [node-fetch](https://github.com/bitinn/node-fetch).
+For a node.js implementation, try [node-fetch](https://github.com/bitinn/node-fetch), or [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch) if you need blob body method.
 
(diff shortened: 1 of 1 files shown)


matthew-andrews:
This seems unlikely…  Isomorphic fetch *is* node fetch

dgraham:
I think the limitations of other packages should be documented in their README files rather than this one. There also seems to be some confusion between the isomorphic-fetch and node-fetch libraries, and not having used them personally, I'm not sure what to do here.

@matthew-andrews If you think this project link is helpful, I can reopen and merge, but I'm closing for now.

mislav:
Isomorphic-fetch seems to have more stars on GitHub. Maybe we could link to both as alternative implementations, but without any commentary, i.e. avoid mentioning that one supports Blob and the other doesn't. The reader is required to do their own research.

tomByrer:
More information would have been helpful for me, more so if there hadn't
been open issues to look though.

mislav:
> More information would have been helpful for me

I understand. We just don't want to maintain any particulars of a feature set for a library we're not responsible for. If node-fetch supported Blob tomorrow, our Readme would be outdated.

PR #317, opened 2016-04-16
refactor Header iterator methods

Xuefeng-Zhu:
remove duplicate code in values, keys, entries method for Header

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -37,6 +37,24 @@
     return value
   }
 
+  // Build a destructive iterator for the value list
+  function iteratorFor(items) {
+    var iterator = {
+      next: function() {
+        var value = items.shift()
+        return {done: value === undefined, value: value}
+      }
+    }
+
+    if (support.iterable) {
+      iterator[Symbol.iterator] = function() {
+        return iterator
+      }
+    }
+
+    return iterator
+  }
+
   function Headers(headers) {
     this.map = {}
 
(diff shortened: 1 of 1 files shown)


fetch.js
```diff
@@ -37,6 +37,25 @@
     return value
   }
 
+  function getIterator(headers, addItem) {
+    var items = []
+    headers.forEach(addItem, items)
```
mislav: Interesting solution to expose `items` as `this` within iteration callbacks

fetch.js
```diff
@@ -93,60 +112,15 @@
...
-    }
-
-    if (support.iterable) {
-      iterator[Symbol.iterator] = function() {
-        return iterator
-      }
-    }
-
-    return iterator
+    return getIterator(this, function(value, name) { this.push(name) })
```
mislav: Cleanup looks good. It's not immediately clear what `this` inside the iteration callback refers to, but that is cleared up when you look up `getIterator` source.

Xuefeng-Zhu: OK. Will add a comment help?

mislav: No I think the comment is too much. If someone is interested in this, whey will inspect the source. It's also now weird that the comment is at the 1st invocation of this function but not on the other ones. Please revert the comment.

Xuefeng-Zhu: OK. I have removed the comment

mislav:
@dgraham This looks like good DRYing up to me. What say you, the author of the original code?

fetch.js
```diff
@@ -37,6 +37,25 @@
...
+    }
+
+    if (support.iterable) {
+      iterator[Symbol.iterator] = function() {
+        return iterator
+      }
+    }
+
+    return iterator
+  }
```
dgraham: We'll get most of the DRY benefit with less complexity if we reduce `getIterator`'s responsibilities to just creating and returning an iterator object over a list of values. It's currently also responsible for building the value list from a function parameter, which complicates each call site.


```js
  // Build a destructive iterator for the value list
  function iteratorFor(items) {
    var iterator = {
      next: function() {
        var value = items.shift()
        return {done: value === undefined, value: value}
      }
    }

    if (support.iterable) {
      iterator[Symbol.iterator] = function() {
        return iterator
      }
    }

    return iterator
  }

// Each call site still builds its own value list.
Headers.prototype.keys = function() {
  var items = []
  this.forEach(function(value, name) { items.push(name) })
  return iteratorFor(items)
}
```

I used `iteratorFor` as the function name in this example because a "get" prefix implies the function returns a saved, pre-calculated value. It's a factory function, though, so maybe `iteratorFor`, `iteratorFrom`, `createIterator`, or simply the noun `iterator` would be better names.

Xuefeng-Zhu: It sounds like a better approach. Do you want me to update the code based on your comment?

Xuefeng-Zhu: I have updated the code

dgraham:
@Xuefeng-Zhu Thanks for your contribution!

PR #331, opened 2016-05-19
Update Webpack section in README

tarikjn:
The currently linked solution for Webpack usage is incorrect as it does not define `fetch` as a global and as a result will fail when an application uses external modules which expect `fetch` to be available, as those will generally not be processed by Webpack's `ProvidePlugin`.

Including `fetch` in the `entry` configuration of Webpack is the correct way to polyfill and fixes this issue. This is consistent with how other polyfills are used e.g. [babel-polyfill](https://babeljs.io/docs/usage/polyfill/), and does not require any special loader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,11 @@ $ npm install whatwg-fetch --save
 
 For a node.js implementation, try [node-fetch](https://github.com/bitinn/node-fetch).
 
-For use with webpack, refer to [Using WebPack with shims and polyfills](http://mts.io/2015/04/08/webpack-shims-polyfills/).
+For use with webpack, add this package in the `entry` configuration option before your application entry point:
+
+```javascript
+entry: ['whatwg-fetch', ...]
+```
 
 For babel and es2015+, make sure to import the file:
 


mislav:
Thanks for the heads-up. Could you simply remove all Webpack documentation instead? We don't want to be updating this every time, and we don't use Webpack ourselves so we can't assess the validity of this over time.

tarikjn:
Up to you, this part of the Webpack API is pretty stable and is identical in the upcoming v2 of Webpack, and it has always been in Webpack, so I don't think it will change much over time.

I can remove this section and post a gist with the instructions. But, the advantage of pointing to the correct instructions is it will help reduce the number of users using the wrong setup. We can also simply point to the [Webpack entry documentation](https://webpack.github.io/docs/configuration.html#entry)

tarikjn:
PS: It looks as if Travis CI is having some issues with some of the tests taking too long and thus showing as failing. This is unlikely to be related to this PR.

mislav:
OK for now I will merge this, but I'm not keen on keeping up with module loader configuration examples if they keep changing. Thanks for the update!

vieks:
@tarikjn Your 'solution' doesn't work at all !
````
npm install whatwg-fetch --save
entry: ['whatwg-fetch', ...]
import 'whatwg-fetch';
````
`'fetch is not defined'`

The old one is OK. Please remove this pull request...

tarikjn:
@vieks you're not supposed to import it when you place it in your entry point. If you still haven't figured it out, please open an issue an mention me, I will also need to look at your full Webpack (or fuller) config so I can help you.

vieks:
@tarikjn Sorry mate, I spoke too fast, in fact both solutions are OK, but as you said your one is the way to go.
My problem was about isomorphic execution, where obviously this 'fetch' implementation is not defined on server side.

Have a good day|night.

PR #335, opened 2016-05-27
Fixes #308. Avoid consuming same body when cloning

### LINKED ISSUE #308: `request.clone()` doesn't preserve the original request
The purpose of the `clone()` method is to allow reuse of the (normally single-use) `Body` of the request. Unfortunately, the implementation in this polyfill doesn't allow that.

To illustrate the problem, this example code will fail:

``` js
var r1 = new Request('http://somewhere.com', { method: 'POST', body: 'test' });
var r2 = r1.clone();
r1.text(); // --> fails - but should succeed
r2.text(); // --> succeeds
```

The `clone()` method in this polyfill just [delegates to the constructor](https://github.com/github/fetch/blob/master/fetch.js#L317-L319), but the constructor is supposed to consume the stream of the original request ([which it does](https://github.com/github/fetch/blob/master/fetch.js#L297)), and the `clone()` method is supposed to leave it untouched.

scarlac:
Body of Request must be cloned using a value to avoid consumption of
it. Once a new body is initialized, re-initialize the old value.
Includes a test that fails with the old code.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -299,9 +299,7 @@
   }
 
   Request.prototype.clone = function() {
-    var newRequest = new Request(this, { body: 'clone' })
-    newRequest._initBody(this._bodyInit)
-    return newRequest
+    return new Request(this, { body: this._bodyInit })
   }
 
   function decode(body) {
--- a/fetch.js
+++ b/fetch.js
@@ -167,6 +167,11 @@
 
     this._initBody = function(body) {
       this._bodyInit = body
+
+      delete this._bodyText
+      delete this._bodyBlob
+      delete this._bodyFormData
+
       if (typeof body === 'string') {
         this._bodyText = body
       } else if (support.blob && Blob.prototype.isPrototypeOf(body)) {
@@ -294,7 +299,9 @@
   }
 
   Request.prototype.clone = function() {
-    return new Request(this)
+    var newRequest = new Request(this, { body: 'clone' })
+    newRequest._initBody(this._bodyInit)
+    return newRequest
   }
 
   function decode(body) {
--- a/fetch.js
+++ b/fetch.js
@@ -168,10 +168,6 @@
     this._initBody = function(body) {
       this._bodyInit = body
 
-      delete this._bodyText
-      delete this._bodyBlob
-      delete this._bodyFormData
-
       if (typeof body === 'string') {
         this._bodyText = body
       } else if (support.blob && Blob.prototype.isPrototypeOf(body)) {
--- a/test/test.js
+++ b/test/test.js
@@ -491,6 +491,19 @@ suite('Request', function() {
     })
   })
 
+  test('clone request bodies', function() {
+    var r1 = new Request('https://fetch.spec.whatwg.org/', {
+      method: 'post',
+      headers: {'content-type': 'text/plain'},
+      body: 'I work out'
+    })
+    var r2 = r1.clone()
+
+    return Promise.all([r1.text(), r2.text()]).then(function(texts){
+      return assert.equal(texts[0], texts[1], 'both requests succeed with equal texts')
+    })
+  })
+
   featureDependent(test, !nativeChrome, 'clone with used Request body', function() {
     var req = new Request('https://fetch.spec.whatwg.org/', {
       method: 'post',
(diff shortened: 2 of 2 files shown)


test/test.js
```diff
@@ -491,6 +491,19 @@ suite('Request', function() {
     })
   })
 
+  test('clone request bodies', function() {
+    var r1 = new Request('https://fetch.spec.whatwg.org/', {
+      method: 'post',
+      headers: {'content-type': 'text/plain'},
```
mislav: I don't think headers are needed for this test.

test/test.js
```diff
@@ -491,6 +491,19 @@ suite('Request', function() {
...
+      body: 'I work out'
+    })
+    var r2 = r1.clone()
+
+    return Promise.all([r1.text(), r2.text()]).then(function(texts){
+      return assert.equal(texts[0], texts[1], 'both requests succeed with equal texts')
```
mislav: I don't think a `return` statement is necessary here.

fetch.js
```diff
@@ -167,6 +167,11 @@
 
     this._initBody = function(body) {
       this._bodyInit = body
+
+      delete this._bodyText
+      delete this._bodyBlob
+      delete this._bodyFormData
```
mislav: Could these deletes be performed in `clone()` instead?

fetch.js
```diff
@@ -294,7 +299,9 @@
   }
 
   Request.prototype.clone = function() {
-    return new Request(this)
+    var newRequest = new Request(this, { body: 'clone' })
```
mislav: How about passing `this._bodyInit` to `body:` parameter instead of `'clone'` which will be discarded anyway?

mislav:
References #308. (In the future, put the reference in issue/PR body instead of title.)

fetch.js
scarlac: That's where I had them initially. I moved them to initBody since it's responsible for setting them. For future maintenance it seemed easier if they are kept close. What do you think is best?

mislav: At second thought, I guess it can't hurt that this step is kept here. And you're right that accessing private properties is better kept close together instead of scattered through multiple function calls.

fetch.js
scarlac: You're right. That would be much simpler. I'll fix that!

test/test.js
```diff
@@ -491,6 +491,19 @@ suite('Request', function() {
     })
   })
 
+  test('clone request bodies', function() {
+    var r1 = new Request('https://fetch.spec.whatwg.org/', {
+      method: 'post',
+      headers: {'content-type': 'text/plain'},
+      body: 'I work out'
```
mislav: Do you think it would be useful to also add a test that verifies that Blob bodies can be cloned too? Take a look at other Blob test to see how to define such tests only for browsers that support Blob.

test/test.js
scarlac: You're right, they're not. I'll remove.

test/test.js
scarlac: I'm unsure since I haven't touched on that part of the code but if we're not polyfilling Blob I'd say it seems a bit unnecessary to test the browser implementation. Did I understand it correctly?

test/test.js
scarlac: You're right again :)

test/test.js
mislav: Well, it wouldn't be testing the browser implementation so much as testing that you can clone a Request with a Blob body and read that body uninterrupted from both clones.

scarlac: Ah I see what you mean. That makes sense to do. I'm writing a test now although it seems to 'just work' so I just need to make sure it's testing properly...

scarlac: I've added a test in the latest commit. It may be that this test would never actually since blobs can be re-read without problems normally but it doesn't hurt to ensure that the clone's body `.blob()` works as well and not just `.text()`.

A more thoroughly test would use a big blob that challenged the reader a bit more but I think that's a bit out of scope for this fix.

fetch.js
mislav: With your latest updates, I don't think these `delete` statements are necessary anymore.

scarlac: I believe you are correct. Since we're now reusing old references, it would always be the same type and thus a delete is unnecessary. I'm removing it now and committing...

PR #341, opened 2016-06-04
Use response.ok in error-handling docs.

yangmillstheory:
Checking the response status is in the 2xx range is equivalent to checking `.ok`.

https://github.com/github/fetch/blob/master/fetch.js#L334

This has the added benefit of documenting that API (`.ok` is a little vague), and making the example code a bit simpler (albeit less explicit).

PS - since this is just a small edit spanning two commits, I think it would be better to squash it in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ status, define a custom response handler:
 
 ```javascript
 function checkStatus(response) {
-  if (response.status >= 200 && response.status < 300) {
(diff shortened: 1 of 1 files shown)


dgraham:
The `ok` attribute is [documented in the specification](https://fetch.spec.whatwg.org/#dom-response-ok). Also see #278.

Issue #343, opened 2016-06-08
ArrayBuffer as body must not be passed directly to XHR


datenreisender:
When `fetch` is invoked with an `ArrayBuffer` as body, that `ArrayBuffer` is passed directly to the `XMLHttpRequest`, but invoking xhr with an `ArrayBuffer` is deprecated. Thus this causes a warning in some browsers, e.g. Safari. It should be wrapped in an `ArrayBufferView`: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send

When executing this fiddle with Safari, the warning is shown in the console: 
https://jsfiddle.net/gkp1swke/4/

datenreisender:
I tried passing an `ArrayBufferView` directly as a body to `fetch`, but this also yields an error, as explained in #344.

inithink:
+1

Issue #344, opened 2016-06-08
Any BufferSource should be accepted as an body for a POST


datenreisender:
According to the spec any BufferSource should be accepted as body for a POST, but this polyfill only allows `ArrayBuffer` not an `ArrayBufferView`. 

Running https://jsfiddle.net/fn2j46tu/1/ on Safari shows in the console, that either using a `DataView` or an `TypedArray` yields a `unsupported BodyInit type` error.

mislav:
Thanks for catching this and providing detail! It seems that there are 2 separate but related problems that you've correctly opened 2 issues for. Let's see if we can tackle them. Would you be willing to submit patches that first address this problem, and then the one described in the other issue?

datenreisender:
I‘m currently not able to submit a tested PR, but my shot would be to utilise [`ArrayBuffer.isView()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/isView) by changing https://github.com/github/fetch/blob/65b611e/fetch.js#L180 to something like 

```js 
  } else if (support.arrayBuffer && (ArrayBuffer.prototype.isPrototypeOf(body) || ArrayBuffer.isView(body))) {
```

PR #351, opened 2016-06-14
Support ArrayBuffer in BodyInit

### LINKED ISSUE #350: BodyInit extract is not supported with arrayBuffer
~~https://github.com/github/fetch/commit/1529c832406dd7f3e78fe6670fae3a29fcdf8c3c added support for arrayBuffer() for POST responses only, but I can't tell why.~~

~~Specifically, in https://github.com/cockroachdb/cockroach/pull/7178 I am relying on arrayBuffer() working in GET responses, which fails under whatwg-fetch but works in Chrome because chrome has native `fetch` support.~~

~~Was there a reason that arrayBuffer() support was omitted for non-POST responses? If not, could you add this support now?~~

EDIT: turns out this is just a simple issue with BodyInit; see https://fetch.spec.whatwg.org/#concept-bodyinit-extract.

tamird:
Fixes #350.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -178,8 +178,12 @@
       } else if (!body) {
         this._bodyText = ''
       } else if (support.arrayBuffer && ArrayBuffer.prototype.isPrototypeOf(body)) {
-        // Only support ArrayBuffers for POST method.
-        // Receiving ArrayBuffers happens via Blobs, instead.
+        if (body.slice) {
+          this._bodyArrayBuffer = new Uint8Array(body.slice(0))
+        } else {
+          this._bodyArrayBuffer = new Uint8Array(body.byteLength)
+          this._bodyArrayBuffer.set(new Uint8Array(body))
+        }
       } else {
         throw new Error('unsupported BodyInit type')
       }
@@ -211,10 +215,6 @@
         }
       }
 
-      this.arrayBuffer = function() {
-        return this.blob().then(readBlobAsArrayBuffer)
-      }
-
       this.text = function() {
         var rejected = consumed(this)
         if (rejected) {
@@ -231,8 +231,17 @@
       }
     } else {
       this.text = function() {
-        var rejected = consumed(this)
-        return rejected ? rejected : Promise.resolve(this._bodyText)
+        return consumed(this) || Promise.resolve(this._bodyText)
+      }
+    }
+
+    if (support.arrayBuffer) {
+      this.arrayBuffer = function() {
+        if (this._bodyArrayBuffer) {
+          return consumed(this) || Promise.resolve(this._bodyArrayBuffer.buffer)
+        } else {
+          return this.blob().then(readBlobAsArrayBuffer)
+        }
       }
     }
 
--- a/test/test.js
+++ b/test/test.js
@@ -444,6 +444,28 @@ suite('Request', function() {
       })
     })
 
+    featureDependent(suite, support.arrayBuffer, 'type ArrayBuffer', function() {
+      test('consume as array buffer', function() {
+        var text = 'name=Hubot'
+
+        var buf = new ArrayBuffer(text.length)
+        var view = new Uint8Array(buf)
+
+        for(var i = 0; i < text.length; i++) {
+          view[i] = text.charCodeAt(i)
+        }
+
+        var request = new Request('', {method: 'POST', body: buf})
+        return request.arrayBuffer().then(function(buffer) {
+          var bufView = new Uint8Array(buffer)
+          assert.equal(bufView.byteLength, view.length)
+          for(var i = 0; i < bufView.length; i++) {
+            assert.equal(bufView[i], view[i])
+          }
+        })
+      })
+    })
+
     suite('type USVString', function() {
       test('consume as text', function() {
         var request = new Request('', {method: 'POST', body: 'hello'})
@@ -478,6 +500,28 @@ suite('Response', function() {
         var response = new Response(new Blob(['hello']))
         return response.text().then(function(text) {
           assert.equal(text, 'hello')
+        })
+      })
+    })
+
+    featureDependent(suite, support.arrayBuffer, 'type ArrayBuffer', function() {
+      test('consume as array buffer', function() {
+        var text = 'name=Hubot'
+
+        var buf = new ArrayBuffer(text.length)
+        var view = new Uint8Array(buf)
+
+        for(var i = 0; i < text.length; i++) {
+          view[i] = text.charCodeAt(i)
+        }
+
+        var response = new Response(buf)
+        return response.arrayBuffer().then(function(buffer) {
+          var bufView = new Uint8Array(buffer)
+          assert.equal(bufView.byteLength, view.byteLength)
+          for(var i = 0; i < bufView.byteLength; i++) {
+            assert.equal(bufView[i], view[i])
+          }
         })
       })
     })


tamird:
Sauce build failed, but the main build is green.

mislav:
Don't worry about Sauce failures for now. But, you should definitely test this manually in browsers that don't support native `fetch()` (such as Safari and IE).

Thanks for the fix. What was the original problem, in your opinion. And, is it possible to simplify the fix? Because, the code seems right now it's doing a lot to just store an ArrayBuffer for body. Where does the complexity come from?

Thanks!

tamird:
Indeed, I've tested this in safari as well as phantomjs.

Regarding the fix, there are two bits of complexity:
1. the spec says to copy the incoming buffer. Ideally, one would use `.slice(0)` for this, but the version of phantomjs that this repo's tests run against doesn't support that method, hence the `if (body.slice)` check and alternative implementation.
2. internally, this library feeds a blob into `Response`'s constructor, so the old way of reading `arrayBuffer()` from `_bodyBlob` must still be supported.

I'm open to suggestions on how to simplify this fix.

mislav:
Should we treat browsers without `ArrayBuffer.slice` support as not supporting ArrayBuffer altogether? Then we can skip the related tests. No sense in providing an alternative implementation that only helps PhantomJS.

tamird:
I don't think the issue is specific to phantomjs, I would expect all older
WebKit browsers to be affected. Why not provide this support? It's pretty
simple to do so.

fetch.js
```diff
@@ -236,6 +236,21 @@
       }
     }
 
+    if (support.arrayBuffer) {
+      this.arrayBuffer = function() {
+        if (this._bodyArrayBuffer) {
+          var rejected = consumed(this)
+          if (rejected) {
+            return rejected
```
mislav: What do you think about condensing this whole condition like:

```js
return consumed(this) || Promise.resolve(...)
```

tamird: I'm fine with this refactor, but this pattern is everywhere in this file. Doesn't seem appropriate to hold this change up over that style debate.

fetch.js
```diff
@@ -236,6 +236,21 @@
...
+          }
+
+          return Promise.resolve(this._bodyArrayBuffer.buffer)
+        } else if (support.blob) {
```
mislav: I wouldn't test for Blob support here. First of all, do any browsers implement ArrayBuffer but not Blob? Second, I don't want this function to ever return `undefined`.

mislav:
Thanks for the explanation. Have you looked into #343 and #344? Do you think those concerns should affect this PR? /cc @datenreisender

fetch.js
mislav: It's fine to not match the style of others in this instance. Other invocations of `consumed()` can not efficiently be condensed into a one-liner — except for `text()`, which already uses a ternary operator.

fetch.js
tamird: Indeed, it seems that `ArrayBuffer` is more widely supported than `Blob`. Fixed this.

fetch.js
tamird: Done.

tamird:
#343 isn't directly related - fixing it requires wrapping before the xhr. #344 is a bit more closely related, but @datenreisender's last suggestion in that thread is already the correct fix - this PR doesn't make that any more difficult.

tamird:
@mislav @datenreisender I've added a commit to fix #343.

tamird:
Sigh, can't reproduce this CI failure locally.

fetch.js
```diff
@@ -426,7 +435,11 @@
         xhr.setRequestHeader(name, value)
       })
 
-      xhr.send(typeof request._bodyInit === 'undefined' ? null : request._bodyInit)
+      if (request._bodyArrayBuffer) {
+        xhr.send(request._bodyArrayBuffer)
```
mislav: Shouldn't we be avoiding passing ArrayBuffer into xhr directly?

mislav: Also, it would be good to make a test for this that performs a real HTTP request (in supporting browsers).

tamird: The tests already exists, or are you suggesting that I test for a deprecation warning? I don't know how to do that.

tamird: The type of `_bodyArrayBuffer` is actually `ArrayBufferView`

mislav: Ah I see. That clears it up; thanks!

tamird:
Updated to try to solve the second commit's test failure.

tamird:
Removed the last commit - passing ArrayBufferView into xhr results in the literal `'[object Uint8Array]'` being sent over the wire.

The version of phantomjs used in CI here is just too old.

tamird:
@mislav I've taken this as far as I can. Will you put me out of my misery? 👼

mislav:
You've helped enough and I thank you. We'll take it over from here. I'll have to look into some things first.

tamird:
@mislav any news here?

mislav:
No, sorry! I need a bit more time.

gaye:
+1

tamird:
@mislav any news?

mislav:
Thanks for rebasing! Haven't gotten around to this yet. Patience!

tmbo:
@mislav any progress here? +1

mislav:
@tamird I want to proceed with this PR, but I need help from you to clarify something.

As far as I understand, these APIs are currently broken in the polyfill, before your changes:

1. A Request instance can be initialized with an ArrayBuffer body, but that body is then not readable via `request.arrayBuffer()` method.
2. Same as (1), but for Response.

Although I agree that these inconsistencies are problematic and need to be fixed, I still don't understand how they interfered with sending & receiving of binary data. These are the facts right now:

1. The `fetch` method **doesn't use** the `request.arrayBuffer()` method to obtain the ArrayBuffer body object. Instead, it reads it directly from `request._bodyInit`. Therefore, sending ArrayBuffers is already possible, and there is a test for that.

2. The `fetch` method **doesn't ever initialize** a Response object with an ArrayBuffer body, because all responses returned from XHR are of type Blob. Therefore, receiving ArrayBuffers from response is already possible, because Blob → ArrayBuffer conversion is functional and achieved through `response.arrayBuffer()`, and there is a test for that too.

Can you help me construct a simple test case that utilizes the `fetch` method to communicate with ArrayBuffers and fails without your fixes? That would be a good starting point for me to get this merged.

tamird:
A great way to ensure that I don't remember the detail is of this is so wait 5 months before asking questions.

We're still using my fork of this library, so it's not as if this has become irrelevant, I just can't remember the details.

> The fetch method doesn't use the request.arrayBuffer() method to obtain the ArrayBuffer body object. Instead, it reads it directly from request._bodyInit. Therefore, sending ArrayBuffers is already possible, and there is a test for that.

This seems patently incorrect based on #350 and the new tests added in this PR. Where is the test you claim exists?

Not sure about your second bullet.

mislav:
> This seems patently incorrect based on #350 and the new tests added in this PR. Where is the test you claim exists?

[It's here](https://github.com/github/fetch/blob/87fd1cd3ad6651ef34e8ba7b667dc19ce0ec865f/test/test.js#L871-L881). It passes even under PhantomJS.

Apologies for sitting on this for so long, but you have to consider my perspective as a maintainer: I have received a pull request that I'm not sure what it solves, has failing tests in CI, and is for a feature that we ourselves don't use at GitHub. This wasn't a really high priority for me.

I'm sure that there is a valid bug somewhere that you discovered, especially since on the cocroachdb project you reported that tests fail on PhantomJS and Safari, which are the browsers where our polyfill is in effect. I just need to first discover how this bug affects `fetch` calls that operate on binary data. I'm working on a more complete ArrayBuffer support, starting with #430.

tamird:
Heh, the failing tests were originally due to sauce labs failing, and after today's rebase were due to a change in the Request constructor, so please, don't try to misrepresent. The "fix" was:
```diff
diff --git a/test/test.js b/test/test.js
index a066e8d..fed12b4 100644
--- a/test/test.js
+++ b/test/test.js
@@ -455,7 +455,7 @@ suite('Request', function() {
           view[i] = text.charCodeAt(i)
         }
 
-        var request = new Request(null, {method: 'POST', body: buf})
+        var request = new Request('', {method: 'POST', body: buf})
         return request.arrayBuffer().then(function(buffer) {
           var bufView = new Uint8Array(buffer)
           assert.equal(bufView.byteLength, view.length)

```

And was needed due to a change you made long after this PR was opened 5f61b12416bc4497f23f67da562b4219b54b49c2

tamird:
I've just confirmed again that moving from my fork to mainline causes tests in CRDB to fail. For what it's worth, here's where we're using this: https://github.com/cockroachdb/cockroach/blob/428c303/pkg/ui/app/util/api.ts#L90:L113

tamird:
Ah! I remember now, this doesn't actually have anything to do with `fetch` itself, it has to do with our use of `fetch-mock` where we need to manually initialize the response object like we do here: https://github.com/cockroachdb/cockroach/blob/428c30331c2ab153450afa110cba6e119e990ec1/pkg/ui/app/util/api.spec.ts#L109:L113

mislav:
> Heh, the failing tests were originally due to sauce labs failing, and after today's rebase were due to a change in the Request constructor, so please, don't try to misrepresent.

Apologies. 🙇 

> I remember now, this doesn't actually have anything to do with `fetch` itself, it has to do with our use of `fetch-mock` where we need to manually initialize the response object

That makes perfect sense. Thanks for looking that up!

tamird:
@mislav where does that leave us? Shall I rebase this again?

mislav:
@tamird No, just leave it to me now. Thanks for the help so far!

tamird:
🎊

mislav:
@tamird Try out the new release and let me know if there are any more problems!

tamird:
We're already using it in CRDB! Thanks again.

PR #358, opened 2016-07-06
Added response to the rejected promise

Teneff:
`xhr.onerror` is triggered when non-success stauts code is returned the response body is still useful


dgraham:
The `onerror` event is triggered by network errors that prevent an HTTP response from arriving. There is no status code, headers, or body available after an error.

PR #375, opened 2016-08-10
Remove an unnecessary trailing comma

jyane:
I have removed an unnecessary trailing comma from README.md.


mislav:
The example is actually valid! Trailing (or "dangling") commas in JS are valid syntax. IE had a problem with those up to version 8, but IE9 and above (and other browsers) handle them quite well. This library doesn't support the old versions of IE, so it's fine.

Issue #383, opened 2016-08-23
X-Request-URL header should be case-insensitive


mruhlin:
If my server returns the x-request-url header in all lowercase, fetch will fail to parse it due (at least in part) to the following case sensitive regex
```
        // Avoid security warnings on getResponseHeader when not allowed by CORS
        if (/^X-Request-URL:/m.test(xhr.getAllResponseHeaders())) {
          return xhr.getResponseHeader('X-Request-URL')
        }
```

per https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html , HTTP headers are case insensitive.

mruhlin:
https://github.com/github/fetch/pull/384

PR #384, opened 2016-08-23
make X-Request-URL header case-insensitive

mruhlin:
HTTP headers are case-insensitive


mruhlin:
fixes https://github.com/github/fetch/issues/383

mislav:
Thank you!

Issue #386, opened 2016-08-28
Response cookies not being set


papigers:
I'm trying to implement client login using fetch on react. 

I'm using passport for authentication. The reason I'm using `fetch` and not regular `form.submit()`, is because I want to be able to recieve error messages from my express server, like: `"username or password is wrong"`.  

I know that passport can send back messages using `flash` messages, but `flash` requires sessions and I would like to avoid them.

This is my code:
```js
    fetch('/login/local', {
          method: 'POST',
          headers: {
            Accept: 'application/json',
            'Content-Type': 'application/json',
          },
          body: JSON.stringify({
            username: this.state.username,
            password: this.state.password,
          }),
        }).then(res => {
          console.log(res.headers.get('set-cookie')); // undefined
          console.log(document.cookie); // nope
          return res.json();
        }).then(json => {
          if (json.success) {
            this.setState({ error: '' });
            this.context.router.push(json.redirect);
          }
          else {
            this.setState({ error: json.error });
          }
        });
```
The server sends the cookies just fine, as you can see on chrome's dev tools:
[![Network - Cookies][1]][1]
[![Network - Headers][2]][2]

But chrome doesn't set the cookies, in Application -> Cookies -> localhost:8080: "The site has no cookies". 

Using `form.submit()` while the server sets the cookies and redirects works just fine, the problem only occurs using fetch to retrieve json, so this is why I'm posting it here. 

Any idea how to make it work?

  [1]: http://i.stack.imgur.com/W8UxQ.png
  [2]: http://i.stack.imgur.com/526e8.png

stianlp:
Having the exact same problem.

From the doc:

> Receiving cookies
> 
> Like with XMLHttpRequest, the Set-Cookie response header returned from the server is a forbidden header name and therefore can't be programatically read with response.headers.get(). Instead, it's the browser's responsibility to handle new cookies being set (if applicable to the current URL). Unless they are HTTP-only, new cookies will be available through document.cookie.
> 

But it doesn't look like my browser is setting the cookies.

stianlp:
So, I've verified that my issue is not caused by the server (like not setting path or somewthing).

I opened chrome://net-internals/#events and looked at the difference between a fetch call and $.ajax call (which sets the cookie correctly).

There is one thing in particular i noticed:

$.ajax:
```

t=587184 [st= 2]   +URL_REQUEST_START_JOB  [dt=6]
[...]

PR #393, opened 2016-09-12
Clarifying the default `Set-Cookie` behavior.

rclmenezes:
I'm trying to make it clear that the `credentials` option affects setting cookies.

Perhaps this will prevent future tickets like https://github.com/github/fetch/issues/386

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ fetch('/avatars', {
 
 ### Caveats
 
-The `fetch` specification differs from `jQuery.ajax()` in mainly two ways that
+The `fetch` specification differs from `jQuery.ajax()` in mainly three ways that
 bear keeping in mind:
 
 * The Promise returned from `fetch()` **won't reject on HTTP error status**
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the suggestion!

Aren't the two pieces of information redundant? They basically state the same thing.

rclmenezes:
I agree that it's a little redundant.

The first section talks about differences between `fetch` and
`jQuery.ajax`. It already mentions that the default behavior of *sending*
cookies is different, so I figure it should mention that receiving cookies
is different as well.

The second section goes more in depth. I just wanted to copy the pattern
about sending cookies.

README.md
```diff
@@ -137,6 +137,8 @@ bear keeping in mind:
 * By default, `fetch` **won't send any cookies** to the server, resulting in
   unauthenticated requests if the site relies on maintaining a user session.
 
+* By default, the **browser will not store any cookies** set in `fetch()` responses.
```
mislav: Since the two bullet points are closely related, could we collapse them into one, explaining how the browser will not handle cookies at all, either sending or receiving?

README.md
```diff
@@ -192,6 +194,9 @@ fetch('https://example.com:1234/users', {
 
 #### Receiving cookies
 
+By default, the browser will ignore the `Set-Cookie` response header unless the
```
mislav: This sentence is good, but please move it 1 paragraph down.

PR #400, opened 2016-09-24
Safari 10 does not have a native implementation

natesilva:
As it turns out, Safari 10 does not support `window.fetch`. This is easily confirmed at the JavaScript console and documented by [caniuse](http://caniuse.com/#feat=fetch). The polyfill code does run on Safari 10.


mislav:
Ouch sorry! I have just amended the README in the response to https://github.com/github/fetch/issues/401 and only after I have seen your PR. Thanks for your contribution, and we'll make sure to pull it in next time! 🙇

Issue #198, opened 2015-08-24
Streaming support


atgillette:
Is there any plans to add support for streaming, something like what [request](https://github.com/request/request) has?

dgraham:
There are no current plans for streaming requests or responses. We are limited by what XMLHttpRequest supports.

There is a [draft specification](https://github.com/yutakahirano/fetch-with-streams) for streams in the Fetch API, with a native implementation [shipping in Chrome](https://www.chromestatus.com/feature/5804334163951616) already.

jimmywarting:
Please reopen, there are two was you can get a stream from XMLHttpRequest.

```javascript
xhrRequest.responseType = "ms-stream";
// or
xhrRequest.responseType = "moz-chunked-arraybuffer";
```

If this would be implemented then i would rather want to use the polyfill then firefox or edge partially supported fetch api
This is crucial for my application where i need to handle large file with lower memory.

mislav:
Sure, let's revisit this. But I make no promises about actually implementing this in github/fetch.

It sounds like we can get native streaming from Chrome 43+, and streaming via XHR in IE and Firefox. How about Safari— is anything available there?

jimmywarting:
Like i said to [jonnyreeves/chunked-request](https://github.com/jonnyreeves/chunked-request/issues/20)
> Played a bit myself with streams and monkey patching xhr: 
This is how i would do it for firefox: https://jsfiddle.net/98L105v3/2/

I created it yesterday, maybe it would be of any interest?

jimmywarting:
Borrow some info from https://github.com/jonnyreeves/chunked-request/issues/16
> - Chrome: use native reader from fetch
- FF: set responseType to moz-chunked-arraybuffer, get ArrayBuffer(s) from xhr.response progress event(s), accumulate if needed, resolve reader.read Promise
- IE10+: set responseType to ms-stream, call readAsArrayBuffer on xhr.response to get an eventemitter, get ArrayBuffer(s) from eventemitter progress event(s), accumulate if needed, resolve reader.read Promise
- IE9: set responseType to text, slice xhr.responseText on progress event(s) to get chunk(s), convert to ArrayBuffer(s) using text-encoding-utf-8, accumulate if needed, resolve reader.read Promise
- Safari: until it supports fetch, basically do the same thing as for IE9, except use native Uint8Array when converting to ArrayBuffer like this.

jimmywarting:
if this gets implemented the we can't use this code any longer
```javascript
[...]

PR #404, opened 2016-10-05
[WIP] Response now accepts ReadableStream as body

jimmywarting:
This is a part necessary to be able to start supporting streaming response (#198)

here is just an example of how to use it

``` javascript
var rs = new ReadableStream({
  start(controller) {
    var chunk = new Uint8Array(3).fill(97)
    controller.enqueue(chunk)
    controller.close()
  }
})

var res = new Response(rs)

res.text().then(console.log) // aaa
res.arrayBuffer().then(console.log) // ArrayBuffer [97, 97, 97]
res.blob().then(console.log) // Blob {size: 3}
var reader = res.body.getReader()
```

This PR don't make use of ReadableStream it just allows end user to be able to construct a Response with a ReadableStream as body as you should be able to do


package.json
```diff
@@ -6,7 +6,7 @@
   "repository": "github/fetch",
   "license": "MIT",
   "devDependencies": {
-    "bower": "1.3.8",
+    "bower": "^1.7.9",
```
mislav: Is this change necessary?

fetch.js
```diff
@@ -204,6 +206,21 @@
...
         if (this._bodyBlob) {
           return Promise.resolve(this._bodyBlob)
+        } else if (this._bodyStream) {
+          var reader = this._bodyStream.getReader();
+          var chunks = [];
+          var pump = function(){
+            return reader.read().then(function(result){
+              if(!result.done){
+                chunks.push(result.value)
+                return pump()
```
mislav: Is it possible to create a `while` loop rather than recursion? I guess not, because of async.

fetch.js
```diff
@@ -204,6 +206,21 @@
 
         if (this._bodyBlob) {
           return Promise.resolve(this._bodyBlob)
+        } else if (this._bodyStream) {
+          var reader = this._bodyStream.getReader();
+          var chunks = [];
+          var pump = function(){
+            return reader.read().then(function(result){
+              if(!result.done){
+                chunks.push(result.value)
```
mislav: Doesn't it defeat the point of streaming if we've just read the entire response into memory at once?

jimmywarting: Mm, no? You call blob, text or arraybuffer so you need to acembly all the chunks into one result.

jimmywarting:
> Thanks, this looks like a good start! Will users of streaming need any polyfills to be able to make use of this? 

They need web-stream-polyfill if they want to use streams. ATM it's like an optional dependency.

> Can you show off some example usage scenario in code?

Can show some example tomorrow. Just realized _bodyStream should just be named body to follow the spec

fetch.js
jimmywarting: This is just for turning a stream into a blob/text/arraybuffer

We don't know in advance if they want a stream/blob/text or a arraybuffer. So I thought the best default responseType should be ms-stream or moz-chunked-arraybuffer instead of blob so we build all our Response method based on streams first

fetch.js
```diff
@@ -177,6 +177,8 @@
         this._bodyText = body.toString()
       } else if (!body) {
         this._bodyText = ''
+      } else if (body.getReader) {
+        this.body = body
```
mislav: Are we now allowing people to interface with `response.body` directly? Is this in the spec?

fetch.js
mislav: Fair point!

package.json
mislav: Deprecation warnings are fine. They don't affect us at all.

fetch.js
jimmywarting: Yep, body is suppose to be a ReadableStream. You can test this directly in chrome's console log

```javascript
rs = new ReadableStream({})
res = new Response(rs)
reader = res.body.getReader()
console.log(reader)

// even this works:
res = new Response(new Blob([]))
reader = res.body.getReader()
console.log(reader)
```

jimmywarting: It's also suppose to solve abortable request by calling `res.body.cancel()` So literally every thing you pass into `new Response` should be transformed to ReadableStream IMO.

But I don't want to rush into things and change the hole polyfill

fetch.js
```diff
@@ -204,6 +206,21 @@
 
         if (this._bodyBlob) {
           return Promise.resolve(this._bodyBlob)
+        } else if (this.body) {
+          var reader = this.body.getReader();
```
mislav: Minor: we don't use semicolons

fetch.js
```diff
@@ -204,6 +206,21 @@
...
+          var chunks = [];
+          var pump = function(){
```
mislav: Minor: we usually put a space in between `)` and `{` (same applies to next line)

fetch.js
```diff
@@ -204,6 +206,21 @@
...
+            return reader.read().then(function(result){
+              if(!result.done){
```
mislav: Minor: we usually put a space after `if`

fetch.js
```diff
@@ -216,6 +233,10 @@
       }
 
       this.text = function() {
+        if (this.body) {
+          return this.blob().then(readBlobAsText)
```
mislav: We have some checks in place (see `consumed()` below) to prevent reading the body data multiple times. Should the same restrictions also apply to readable streams?

jimmywarting:
Will try to add some test and fix the minor issues.
The stream api already has some sort of consumed restriction in a way that you can not call body.getReader() twice...

jimmywarting:
Just discovered this:
```javascript
rs = new ReadableStream()
res = new Response(rs)
reader = res.body.getReader()
console.log(res.bodyUsed) // false
reader.read()
console.log(res.bodyUsed) // true
```

jimmywarting:
Think I'm going to need a little bit of help with the test. Got a error: <br>`Uncaught Error: Out of skill VM397:1`

mislav:
Not sure what that error is about, but PhantomJS is definitely going to need some ES shims before it can load the streams polyfill. PhantomJS has neither Symbol nor Object.assign support. Your test, however, runs but fails in Chrome with a different error. Perhaps you should work on making the test work in a real browser, and we can handle eventually getting it to run in PhantomJS.

mislav:
@jimmywarting Here are [my polyfills for PhantomJS](https://github.com/mislav/fetch/compare/cd3c6ff6bd604f19449134c7b578d8ff03f678cb...streams). It still fails with a cryptic exception that's different than yours.

jimmywarting:
Works in the browser now...

jimmywarting:
Weird, running `make test` locally yields
```
  91 passing (196ms)
  8 pending
  0 failure 
```

even the browser succeeded

jimmywarting:
Do you have any idea how to tackle this? The test passes in browser but not in phantomjs

mislav:
Not sure. We could just exclude this test from PhantomJS then. But that's a shame, since this test won't ever get exercised in CI, and we risk this functionality falling out of date during future changes.

jimmywarting:
This looks good to me now...

jimmywarting:
> It adds the response.body property conditionally, meaning it won't be present unless the response is streamable. I have to check the spec about this, but I kind of feel that this property should always be available.

I would like response.body to always be there, But that would mean that this fetch polyfill will need to be dependent on some ReadableStream polyfill - I would be fine with that cuz that is what I'm after and why I'm trying to help you implement this.
If it was included then we could get rid of `_bodyText, _bodyBlob, _bodyFormData` and everything else since the bodyInit would always have to be transformed into a ReadableStream

fetch.js
```diff
@@ -341,7 +370,19 @@
   Body.call(Response.prototype)
 
   Response.prototype.clone = function() {
-    return new Response(this._bodyInit, {
+    var clone
+
+    if (this.bodyUsed) {
+      throw new TypeError('Failed to execute \'clone\' on \'Response\': Response body is already used')
```
jimmywarting: I used it first but then the linter complained about using 2 different quotes...

fetch.js
```diff
@@ -177,6 +194,14 @@
         this._bodyText = body.toString()
       } else if (!body) {
         this._bodyText = ''
+      } else if (body.getReader) {
+        this.body = body
+        Object.defineProperty(this, 'bodyUsed', {
```
jimmywarting: I wasn't a big fan of this either, I even asked the community if it was okey to do this: https://github.com/whatwg/fetch/issues/396

But it wasn't... We should maybe implement a new TypeError message that checks the locked property. 
It's a bit more complicated to set the bodyUsed property since it needs set it upon the first getReader().read call and spying on this is not that easy...

fetch.js
jimmywarting: Could as well fall thought and use tee()'s error message that already checks if stream is locked

test/test.js
```diff
@@ -577,6 +586,15 @@ suite('Response', function() {
     })
   })
 
+  // Firefox & Edge partially support fetch (but don't have stream support)
+  // However the polyfilled Response do support it (With web-streams-polyfill)
+  // PhantomJS can use the polyfill but failes togheter with mocha
+  featureDependent(test, !isPhantomJS && support.stream, 'Response should accept ReadableStream', function() {
```
jimmywarting: done

eko24:
@mislav any chances to see this changes in `master` ?

mislav:
@eko24 Not at the present moment. I'm mostly away from my computer around the New Year and I have to re-review this once I get back. We definitely want to support this, but I need to study the spec around this myself and see if we can minimize the diff further. The soonest I might be able to get back to this is 2 weeks from now.

eko24:
Ok, thank you for update !

вт, 27 дек. 2016 г. в 16:48, Mislav Marohnić <notifications@github.com>:

boczeratul:
hello!
any update on this one?

mislav:
@jimmywarting Thank you for your contribution and sorry that we have never reached a resolution regarding this! I have felt uneasy about all the code added to support this feature, which might prove to be tricky to maintain in the future and also I don't really see it being widely used. I'm calling this polyfill basically feature-frozen now https://github.com/github/fetch/issues/198#issuecomment-494361060

jimmywarting:
roger that

PR #408, opened 2016-10-14
Fix grammar

marciopuga:
"similar" -> "similarly"

The adjective similar is modifying behave instead of a noun or pronoun. 

> Use an adverb to modify a verb, adjective, or another adverb.


mislav:
Good eye; thanks!

Issue #355, opened 2016-06-24
Nexus 7 Dolphin 11 "Object #<Response> has no method 'blob'"


tobireif:
On Nexus 7 in Dolphin 11 I'm getting "Object #<Response> has no method 'blob'".

tobireif:
Also in Android Browser 4.1.

dgraham:
The [Response.blob](https://github.com/github/fetch/blob/4bde61788365cfd3d6155eee8aed3c7673d299d5/fetch.js#L199) function is available only when [Blob is supported](https://github.com/github/fetch/blob/4bde61788365cfd3d6155eee8aed3c7673d299d5/fetch.js#L11) by the browser. The Blob constructor was first [available on Android](http://caniuse.com/#feat=blobbuilder) with Chrome 50.

tobireif:
Thanks!

For future visitors, here's the feature-detection code you linked:
```
'FileReader' in self && 'Blob' in self && (function() {
  try {
    new Blob()
    return true
  } catch(e) {
    return false
  }
})()
```

jimmywarting:
I think it's insufficient to only include the blob method when it's possible to use it. Leaving the user with question to why it doesn't exist.

`res.blob()` returns a promise and I think we could use it and reject if it fails to construct a blob. With a message like 

```javascript
reject(new Error("Failed to construct a blob, Include blob a polyfill, https://github.com/eligrey/Blob.js"))
```

Leaving us with fewer browser specific hacks

mislav:
@jimmywarting That's an excellent idea. We'd love a PR that does that.

PR #410, opened 2016-10-16
Throw error when unable to construct a blob #355

jimmywarting:
also makes arraybuffer() always avalible

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -8,14 +8,6 @@
   var support = {
     searchParams: 'URLSearchParams' in self,
     iterable: 'Symbol' in self && 'iterator' in Symbol,
-    blob: 'FileReader' in self && 'Blob' in self && (function() {
-      try {
-        new Blob()
-        return true
-      } catch(e) {
-        return false
-      }
-    })(),
     formData: 'FormData' in self,
     arrayBuffer: 'ArrayBuffer' in self
   }
@@ -169,7 +161,7 @@
       this._bodyInit = body
       if (typeof body === 'string') {
         this._bodyText = body
-      } else if (support.blob && Blob.prototype.isPrototypeOf(body)) {
+      } else if (self.Blob && Blob.prototype.isPrototypeOf(body)) {
         this._bodyBlob = body
       } else if (support.formData && FormData.prototype.isPrototypeOf(body)) {
         this._bodyFormData = body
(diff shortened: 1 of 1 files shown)


fetch.js
```diff
@@ -418,9 +408,7 @@
         xhr.withCredentials = true
       }
 
-      if ('responseType' in xhr && support.blob) {
```
mislav: There should still be a conditional here.

fetch.js
```diff
@@ -8,14 +8,6 @@
   var support = {
     searchParams: 'URLSearchParams' in self,
     iterable: 'Symbol' in self && 'iterator' in Symbol,
-    blob: 'FileReader' in self && 'Blob' in self && (function() {
```
mislav: I vote that we keep this feature test instead of checking for `self.Blob` presence. It's more comprehensive.

fetch.js
```diff
@@ -169,7 +161,7 @@
       this._bodyInit = body
       if (typeof body === 'string') {
         this._bodyText = body
-      } else if (support.blob && Blob.prototype.isPrototypeOf(body)) {
+      } else if (self.Blob && Blob.prototype.isPrototypeOf(body)) {
```
mislav: Ditto my last comment. `support.blob` is better here.

fetch.js
```diff
@@ -195,44 +187,42 @@
...
-        return this.blob().then(readBlobAsArrayBuffer)
+      if (this._bodyBlob) {
+        return Promise.resolve(this._bodyBlob)
+      } else if (this._bodyFormData) {
+        throw new Error('could not read FormData body as blob')
+      } else {
+        try {
+          return Promise.resolve(new Blob([this._bodyText]))
+        } catch(e) {
+          throw new Error('Failed to construct a Blob, Include a Blob polyfill')
```
mislav: I don't like this being a catch-all block. It might swallow exceptions that we didn't intend. Instead, simply leave `Promise.resolve(new Blob(...))` unwrapped and let `new Blob` raise a "'Blob' is undefined" ReferenceError.

fetch.js
```diff
@@ -195,44 +187,42 @@
...
+    }
 
-        if (this._bodyBlob) {
-          return readBlobAsText(this._bodyBlob)
-        } else if (this._bodyFormData) {
-          throw new Error('could not read FormData body as text')
-        } else {
-          return Promise.resolve(this._bodyText)
-        }
+    this.text = function() {
```
mislav: I like this unconditional definition of `text()` much better than previous code! 👍

dgraham:
The previous code defined the `blob` and `arrayBuffer` methods only if the browser supports the `Blob` API. Attempting to read the response as a blob in Internet Explorer, for example, would result in a method missing error. This branch changes the response objects to always include a `blob` method and reject the promise when the browser doesn't support the `Blob` type.

#### Before

```js
fetch('/').then(response => response.blob()) // blob method missing error
```

#### After

```js
fetch('/').then(response => response.blob()).catch(error => …) // Blob is undefined error
```

It makes sense that `Response` would always provide the same set of methods and return rejected promises if the browser doesn't support enough API to properly implement them.

Anyone relying on the old behavior might have this feature test:

```js
if (response.blob) {
  response.blob().then(…)
} else {
  // text()?
}
```

That would need to move to the catch handler:

```js
  response.blob().then(…).catch(error => /* try again with text() */)
```

jimmywarting:
That pretty much sums it up

fetch.js
```diff
@@ -195,44 +195,40 @@
...
 
-      this.arrayBuffer = function() {
-        return this.blob().then(readBlobAsArrayBuffer)
+      if (this._bodyBlob) {
+        return Promise.resolve(this._bodyBlob)
+      } else if (this._bodyFormData) {
+        throw new Error('could not read FormData body as blob')
+      } else {
+        return new Promise(function(resolve){
+          resolve(new Blob([this._bodyText]))
```
mislav: I was happier with the previous approach. I expected the `blob()` method to throw `Blob is not defined`, not reject.

jimmywarting: Potato tomato when i wrote this i thought rejection ^^
I thought yea, it should always return a Promise... You must handle consumed either way also

When you are writing it like this it will end up doing the same thing
```javascript
fetch('/').then(response => response.blob()).catch(error => …) // Blob is undefined error
```

but what if you wrote it like this
```javascript
res = new Response('x')
res.text()
try {
res.blob().catch(err => {
  // body have been used
})
} catch(err) {
  // blob not supported
}
```

mislav:
Hi @jimmywarting, I'm going to close this out after all. Even though I was personally in favor of this cleanup, my teammates have convinced me that conditionally defining these methods makes feature detection easier for the user, i.e. they can detect whether `blob` exists before invoking `blob()` function.

I know it was I who asked you to submit this PR initially, but now it seems this won't be the direction we'll be taking. Sorry for that, and thanks for your patience!

PR #419, opened 2016-10-25
Improve the example with headers in the README

Angelmmiguel:
I found some problems running the POST JSON example in my project. I use a modern Chrome version, so it uses a native `fetch` implementation. In this case, if I send the headers like the example, I'd use an object:

```javascript
var headers = {
  'Accept': 'application/json',
  'Content-Type': 'application/json'
}
```

However, this object is not valid for native `fetch` because it requires a `Headers` instance. I found this in [Using Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Headers) documentation from Mozilla. After update my code to use `Headers`, it works fine:

```javascript
var headers =  new Headers({
  'Accept': 'application/json',
  'Content-Type': 'application/json'
})
```

I checked your code and I saw that it's compatible with `Headers` object ([fetch.js#L61](https://github.com/github/fetch/blob/master/fetch.js#L61)).

Thanks! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,10 +98,10 @@ fetch('/users', {
 ```javascript
 fetch('/users', {
   method: 'POST',
-  headers: {
+  headers: new Headers({
     'Accept': 'application/json',
     'Content-Type': 'application/json'
-  },
+  }),
   body: JSON.stringify({
     name: 'Hubot',
     login: 'hubot',


mislav:
The `fetch()` function definitely accepts a plain object for the `headers:` parameter. It says so [in the MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch/fetch). Can you share with us the code that is failing in Chrome?

Angelmmiguel:
Sorry @mislav, you are right. I think I had some sync problem between my source code and the browser. It works with an object or a `Headers` instance.

Btw, I'm going to switch to `Headers` because I think it's more descriptive. Feel free to close the PR if you want to use a plain object as default for the header object :)

mislav:
Yes we want to use the plain object in examples to keep them concise. Thanks for writing in!

PR #434, opened 2016-11-15
Make it clear what the promise will catch

freeall:


joshuakarjala:
Or just remove any attempt at explaining :)

mislav:
@freeall Thanks for the suggestion. Although you are technically correct about the final `catch` block handling failures that might be unrelated to JSON, our examples are provided for simplicity, and we don't plan to cover all the nuances of promise-based exception handling. Therefore, I feel that this example is fine if left unchanged.

I'm open to accepting a separate example that shows how to handle failures coming directly from `fetch` call separately from handling exceptions that might have happened in any processing of the response afterwards.

freeall:
@mislav How would you actually do that? Is it possible to check the `ex` (err?) and see which error it was?

mislav:
@freeall 

```js
fetch(...).then(function(result) {
  // do whatever with result
}, function(err) {
  // this handles network errors
}).then(function() {
  // do whatever with result
}).catch(function(result) {
  // handle all other errors
})
```

Issue #405, opened 2016-10-07
Really needs a timeout option


john-doherty:
This library needs the ability to provide a timeout interval as part of the options object:

```js
fetch('/users', {
  method: 'POST',
  timeout: 1000,
  headers: {
    'Accept': 'application/json',
    'Content-Type': 'application/json'
  },
  body: JSON.stringify({
    name: 'Hubot',
    login: 'hubot',
  })
})
```

which is then internally assigned to the [xhr](https://github.com/github/fetch/blob/master/fetch.js#L381) object ```xhr.timeout = 1000``` _(property details [here](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/timeout))_.

I'm having to do [this](https://gist.github.com/john-doherty/bcf35d39d8b30d01ae51ccdecf6c94f5), which feels over-the-top:

```js
function promiseTimeout(ms, promise){

  return new Promise(function(resolve, reject){

    // create a timeout to reject promise if not resolved
    var timer = setTimeout(function(){
        reject(new Error("promise timeout"));
    }, ms);

    promise
        .then(function(res){
            clearTimeout(timer);
            resolve(res);
        })
        .catch(function(err){
            clearTimeout(timer);
            reject(err);
        });
  });
};
```

```js
promiseTimeout(1000, fetch('https://courseof.life/johndoherty.json'))
   .then(function(cvData){
      alert(cvData);
   })
   .catch(function(){
      alert('request either failed or timedout');
   });
```

mislav:
Yes you're doing it right: https://github.com/github/fetch/issues/175#issuecomment-125779262

We can't add a timeout option since it's not in the spec.

DinoSourcesRex:
Can you...add it to the spec?

john-doherty:
I built a fetch wrapper that adds offline functionality; it includes timeout/retry options:

```js
var options = {

    // typical fetch property
    method: 'GET',

    // new offline property with optional config
    offline: {
        storage: 'localStorage',    // use localStorage (defaults to sessionStorage)
        timeout: 750,               // request timeout in milliseconds, defaults 730ms
        expires: 1000,              // expires in milliseconds, defaults 1000ms (set to -1 to check for updates with every request)
        debug: true,                // console log request info to help with debugging
        renew: false,               // if true, request is fetched regardless of expire state. Response is and added to cache

        // timeouts are not retried as they risk cause the browser to hang
[...]

Issue #246, opened 2015-12-10
Timeout for request failure


jnatherley:
Is it currently possible to set the maximum amount of retries and also the duration before marking a request as failed?

dgraham:
See #16 and #175 for a discussion on timeouts.

Issue #181, opened 2015-07-28
How to set Authorization and Timeout 


sunnybogawat:
I am doing a get request and I need to pass the authorization token in the request also I need to override the timeout setting. I haven't found any way to do that in the documentation.

dgraham:
See [this section](https://github.com/github/fetch#post-json) of the readme for header syntax. See #175 for timeout discussion.

sunnybogawat:
I am following the same way but when I debugged the ajax request I am not able to see the authorization token. I am doing cross origin request. 


fetch('/users', {
  method: 'post',
  headers: {
    'Accept': 'application/json',
    'Content-Type': 'application/json',
    'Authorization': 'Token token=xxxxx'
  }})

dgraham:
The Chrome Network panel can help debug cross-origin requests.

First, disable the cache so the OPTIONS pre-flight request will occur for each cross-origin request.

![screenshot 2015-07-29 10 51 19](https://cloud.githubusercontent.com/assets/122102/8964010/6aee8ac4-35e0-11e5-9989-242c427ef021.png)

Then in the Network panel you can see both the OPTIONS pre-flight and the subsequent actual request. It's likely that OPTIONS is failing so your request is never made.

![screen shot 2015-07-29 at 10 57 06 am](https://cloud.githubusercontent.com/assets/122102/8964045/96989cd2-35e0-11e5-82f5-6028c4b02a87.png)

sunnybogawat:
Thanks for a quick reply. I have the same setting in my chrome agent and I am passing the token to the fetch request also but somehow it failing at OPTIONS request.  Please find below screen shot to understand the workflow 

1. The Fetch Request Code - 

![capture2](https://cloud.githubusercontent.com/assets/250483/8976870/a08160aa-36b0-11e5-8f41-3ff5583fc8cb.PNG)

2. My Chrome Setting

![capture](https://cloud.githubusercontent.com/assets/250483/8976903/d6e3f158-36b0-11e5-8b9d-767900f0ee8b.PNG)

3. Get Request Created and Error Received With Request Header

![2015-07-30_1140](https://cloud.githubusercontent.com/assets/250483/8976911/f4da5c42-36b0-11e5-8efa-e6d0438ab8d6.png)


![2015-07-30_1139](https://cloud.githubusercontent.com/assets/250483/8976912/f4dd00e6-36b0-11e5-9b8c-684222d2e9ef.png)


[...]

Issue #175, opened 2015-07-14
[Q] timeout handling


gaydenko:
Hi! 

Haven't found the way to set connection timeout. How to?

dgraham:
There is an open discussion on timeout handling in the Fetch specification repository: https://github.com/whatwg/fetch/issues/20. We will add the behavior to the polyfill when the feature is added to the standard specification.

mislav:
You can implement your own timeout wrapper for Promises:

```js
// Rough implementation. Untested.
function timeout(ms, promise) {
  return new Promise(function(resolve, reject) {
    setTimeout(function() {
      reject(new Error("timeout"))
    }, ms)
    promise.then(resolve, reject)
  })
}

timeout(1000, fetch('/hello')).then(function(response) {
  // process response
}).catch(function(error) {
  // might be a timeout error
})
```

The `timeout()` function will wrap any promise and ensure that it's at least rejected within `ms` milliseconds. If the `fetch` succeeds to resolve earlier than that, then the request will be successful.

Note that this is not a connection timeout. This is a response timeout. Due to technical restrictions we can't implement a connection timeout. Also note that with the above implementation, even if the timeout happends the original request won't be aborted because we don't have an API to abort `fetch` requests. Instead, the request will complete in the background but its response will get discarded.

deju:
@mislav  Awesome!

srahulprdxn:
@mislav Good one.

nodkz:
Five cents to @mislav solution - remove timeout if promise was resolver or rejected internally. 

```js
function timeoutPromise(ms, promise) {
  return new Promise((resolve, reject) => {
    const timeoutId = setTimeout(() => {
      reject(new Error("promise timeout"))
    }, ms);
    promise.then(
      (res) => {
        clearTimeout(timeoutId);
        resolve(res);
      },
      (err) => {
        clearTimeout(timeoutId);
        reject(err);
      }
    );
  })
}
```

L-Jovi:
@nodkz  @mislav  thanks for your idea!

and how could I use it in ES7 async/await syntax?

L-Jovi:
I ensure above statement can't suitable for ES7 async/await syntax,
but I don't know how to call `timeout` method, any idea?

gre:
i wish this was actually possible to stop the underlying fetch() work...
shortcuting the flow but not interrupting is not really satisfying.
I really wish fetch() was interruptable like in `xhr.abort()`..

mislav:
[...]

PR #435, opened 2016-11-16
added timeout support

Gcaufy:
it's really useful to add a timeout option.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -330,6 +330,7 @@
     }
     this.method = normalizeMethod(options.method || this.method || 'GET')
     this.mode = options.mode || this.mode || null
+    this.timeout = options.timeout || 0
     this.referrer = null
 
     if ((this.method === 'GET' || this.method === 'HEAD') && body) {
@@ -420,6 +421,9 @@
       var request = new Request(input, init)
       var xhr = new XMLHttpRequest()
 
+      if (request.timeout)
+        xhr.timeout = request.timeout
+
       xhr.onload = function() {
         var options = {
           status: xhr.status,
@@ -436,6 +440,7 @@
       }
 
       xhr.ontimeout = function() {
+        xhr.abort();
         reject(new TypeError('Network request failed'))
       }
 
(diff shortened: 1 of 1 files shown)


mislav:
Thanks, but we can't accept a feature that's not in the spec! See https://github.com/github/fetch#read-this-first and https://github.com/github/fetch/issues/175#issuecomment-125779262 https://github.com/github/fetch/issues/405 https://github.com/github/fetch/issues/246 https://github.com/github/fetch/issues/181 https://github.com/github/fetch/pull/68 https://github.com/github/fetch/issues/16 for background.

Gcaufy:
ok, got it.

fritx:
Excuse me, how could we increase timeout to Infinity or over 120s which is default value of Chrome?

I couldn't find out a solution from those issues :/

Gcaufy:
you can use Promise.race instead.

```
const fetchTimeout = (p, time) => {
    let timeout = new Promise((resolve, reject) => { 
        setTimeout(() => reject(new Error('timeout')), time); 
    });
    return Promise.race([p, timeout]);
};


fetchTimeout(fetch({}), 3000).then()
```

fritx:
@Gcaufy your code looks like to declare a timeout instead, but I want to drop any timeout that would exists. Am I missing anything?

Gcaufy:
this is like 
```
axios({timeout: 3000})
```

so I guess what you need is
```
axios.defaults.timeout = 3000
```
or a common call back, like: 
```
fetch.globalTimeoutCallback = () => {
   console.log('this is timeout');
}
```

what ever you need, you can use Promise.race get what you want:
e.g.
```
const reqest = (p, time = null) => {
   time = time || fetch.defaults.timeout;
   if (!time) return p;
   let timer;
   let timePromise = new Promise((resolve, reject) => {
     timer = setTimeout(() => {
        fetch.globalTimeoutCallback();
        reject(new Error('timeout'));
    }, time)
   });
   return Promise.race([p.then(res => {
        clearInterval(timer);
        return res;
    ), timer])
}
```

jimmywarting:
actually, this is what you should be doing:

```js
const controller = new AbortController()
const signal = controller.signal
setTimeout(() => { 
  controller.abort()
}, 1000)

fetch(url, { signal })
```

PR #440, opened 2016-11-19
Remove incorrect set of bodyUsed flag when cloning a request

morsdyce:
This removes setting bodyUsed to true when you clone a request.

In order to use a request multiple times you need to clone a request, which lets you get a copy of the request while leaving the original request as is, as long as you didn't read the request body beforehand.

In Chrome implementation for example cloning a request does not change the bodyUsed property to true, however in this polyfill it does.

This prevents you from cloning a request and passing on the original request as passing it causes the Already Read exception even though the body was not read directly from that request (by the end user).

**Polyfill scenario:**
https://runkit.com/58308876f439290014bb0025/5830907cf836f700146fa4f7

**Chrome Implementation scenario:** (open the browser console to see the output)
https://plnkr.co/edit/OVK9RREqhMm3WcgPtxOf?p=preview

Sorry for not having a single place for both scenarios. I could not find a snippet service that allows me to bring the package from npm.

**TL;DR**
bodyUsed should not be set to true when cloning a request on the original request

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -299,7 +299,7 @@
     return (methods.indexOf(upcased) > -1) ? upcased : method
   }
 
-  function Request(input, options) {
+  function Request(input, options, isCloneOperation) {
     options = options || {}
     var body = options.body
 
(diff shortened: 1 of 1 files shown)


mislav:
Thanks, but if you tried to add a failing test before coding up a solution, you would have found out that `bodyUsed` is not actually set to true. We have tests to verify that behavior as well. We only had tests to verify requests with bodies, though, so I've added a new test that verifies this behavior for GET request too, like in your runkit example.

PR #448, opened 2016-12-03
Replace catch in examples with `then` error handler

SBoudrias:
Suggesting to use `.catch()` in your examples lead newer developper to mislead potential code errors for a network failures.

This is pretty bad because it hides away legitimate code errors we'd want to catch and logs for future fixes (or immediate developper attention in the devtool console).

By using the `then(onSuccess, onError)`, we're "sure" the error handler is triggered when a network failure occurred. Using `.catch()`, you'll actually be handling network failure AND any exception raised inside the `.then()` success handler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ fetch('/users.json')
     return response.json()
   }).then(function(json) {
     console.log('parsed json', json)
-  }).catch(function(ex) {
+  }, function(ex) {
     console.log('parsing failed', ex)
   })
 ```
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for suggesting. However, this was suggested before, and I denied it on the grounds of not significantly improving the documentation, and instead making it slightly less readable https://github.com/github/fetch/pull/434#issuecomment-260685301

Our examples are meant to be understandable at a glance, not an example of production code. It's the responsibility of the reader to familiarize themselves with the nuances of promise-based error handling.

Even if we make the change you're suggesting, that's still not 100% precise. The only way to handle network errors and *only* network errors is as an error callback directly on the first `fetch()` call, before any chaining happened. After you started chaining `then()` methods, any error callbacks added later will handle both network errors and possible errors/rejections in any of the callbacks up to this point.

However, I don't see the value in expanding our examples with meticulous error handling, since that might make them appear convoluted.

Issue #422, opened 2016-10-28
Handling of server header responses with carriage return \r (IE11)


opinel:
Hi,

I am seeing an issue with the fetch.js method when extracting header values in IE11.
My server response headers have a trailing carriage return "\r".
In Safari, the header value pairs are extracted correctly:

`var pairs = (xhr.getAllResponseHeaders() || '').trim().split('\n')`

However, in IE11, it considers the carriage return as a pair, too. Therefore it runs into an exception case when calling `normalizeName()` within the `append()` method, complaining that the header would include an invalid character `"Invalid character in header field name"`.

Has someone experienced the same problem? And could be potentially filter carriage returns proactively?

mislav:
Hi thanks for reporting. It's very weird that we chose to split headers on `\n`, because the HTTP spec dictates that headers should be split by `\r\n` always. Could you perhaps take a shot at making a test that demonstrates this failure in IE and submit a fix?

PR #449, opened 2016-12-05
Support Tolerance Provision when parsing headers

woudini:
The change to fix this issue https://github.com/github/fetch/issues/422 has created another issue for our app on devices running iOS 8.2 and iOS 9. The impact prevents these users from using our app entirely.

More specifically, the result of splitting by `\r\n` and not both `\r\n` and `\n` is a malformed response header with all the actual headers put into the 'pragma' key and a content type of `text/plain;charset=UTF-8` instead of `application/json`.

Source:
http://stackoverflow.com/a/5757349
more specifically, https://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html#sec19.3

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -355,7 +355,7 @@
 
   function parseHeaders(rawHeaders) {
     var headers = new Headers()
-    rawHeaders.split('\r\n').forEach(function(line) {
+    rawHeaders.split(/\r?\n/).forEach(function(line) {
       var parts = line.split(':')
       var key = parts.shift().trim()
       if (key) {


mislav:
Thanks for reporting. Why support the single `\r`, when the tolerance provision doesn't specify it?

woudini:
Changed to remove support for single `\r`.

fetch.js
```diff
@@ -355,7 +355,7 @@
 
   function parseHeaders(rawHeaders) {
     var headers = new Headers()
-    rawHeaders.split('\r\n').forEach(function(line) {
+    rawHeaders.replace('\r\n', '\n').split('\n').forEach(function(line) {
```
mislav: Would this work as simply `split(/\r?\n/)` to avoid the replace?

mpfluger: Yes, that works - thanks for your quick response!

woudini:
Thanks!

woudini:
Edited scope of impact in original post.

PR #464, opened 2017-01-14
When I build by rollup, I got the bellow error.

kwmt:
DON'T merge. The PR that I only want to watch this diff.

⚠️   The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten
https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
node_modules/whatwg-fetch/fetch.js (460:40)
458:   }
459:   self.fetch.polyfill = true
460: })(typeof self !== 'undefined' ? self : this);
                                             ^

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -455,4 +455,6 @@
     })
   }
   self.fetch.polyfill = true
-})(typeof self !== 'undefined' ? self : this);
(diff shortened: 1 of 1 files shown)


kwmt:
Sorry, I mistook.

PR #474, opened 2017-02-11
Getting fetch support for onreadystatechange

lygstate:


mislav:
Hello; can you explain more about why you think this change is necessary?

We tend to not accept pull requests without any explanation. Thanks!

lygstate:
On some enviornment such As Qt doesn't support for onload but support for onreadystatechange
And it's support for ArrayBuffer but doesn't support for Blob.

dgraham:
The `onload` handler has been working well in production for our supported browser list for a couple years, so I'm reluctant to make this change now.

PR #484, opened 2017-02-21
Point to GitHub Releases in CHANGELOG.md

strugee:


mislav:
Thanks for the suggestion, but I'd rather not have a CHANGELOG file if it's going to always be empty. I think the presence of a CHANGELOG file should always mean that there is information inside it.

strugee:
@mislav would you take a PR that copied GitHub release information into there, then?

The fundamental problem here is that `CHANGELOG.md` is a very, very standard place to look for this information. GitHub Releases is not (yet?). Lots of people won't be able to find this information and will be annoyed for no reason. (Honestly that seems like a good enough reason to accept this... what exactly is the concrete problem with having an "empty" file?)

mislav:
> would you take a PR that copied GitHub release information into there, then?

No, because I would have to take such PRs after every new release. The Releases section is where the release notes are, and nowhere else.

> Lots of people won't be able to find this information and will be annoyed for no reason.

I agree that it's not immediately obvious that this information is in the Releases section. Would an acceptable solution be to point that out in the README?

strugee:
@mislav that would definitely be _way_ better than nothing! But I still don't quite understand why adding this file is such a big deal - that seems like the optimal solution to me.

PR #485, opened 2017-02-22
Accept array in Headers constructor

tschaub:
[From the spec](https://fetch.spec.whatwg.org/#typedefdef-headersinit), it looks like the `Headers` constructor should accept an array of arrays.

I've adapted [the example](https://fetch.spec.whatwg.org/#example-headers-class) for the tests:
```js
var meta = { "Content-Type": "text/xml", "Breaking-Bad": "<3" }
new Headers(meta)

// The above is equivalent to
var meta = [
  [ "Content-Type", "text/xml" ],
  [ "Breaking-Bad", "<3" ]
]
new Headers(meta)
```

This came up for me as an issue using `stream-http` in Edge with the Financial Times polyfill.  I've opened jhiesey/stream-http#71 and Financial-Times/polyfill-service#1091 in case this can be addressed elsewhere, but it looks to me like this might be the right fix.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -84,7 +84,10 @@
       headers.forEach(function(value, name) {
         this.append(name, value)
       }, this)
-
+    } else if (Array.isArray(headers)) {
+      headers.forEach(function(header) {
+        this.append(header[0], header[1]);
+      }, this);
     } else if (headers) {
       Object.getOwnPropertyNames(headers).forEach(function(name) {
         this.append(name, headers[name])
--- a/test/test.js
+++ b/test/test.js
@@ -170,6 +170,16 @@ suite('Headers', function() {
     assert.equal(headers.get('Accept'), 'application/json,text/plain')
     assert.equal(headers.get('Content-type'), 'text/html')
   })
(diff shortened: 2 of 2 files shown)


dgraham:
The native `Headers` implementation in Chrome, Firefox, and Edge accept an array in the constructor. IE 11 and Safari 10, where this polyfill is active, does not, so this is indeed a bug in the Headers polyfill.

Thanks for the detailed bug report and fix.

fetch.js
```diff
@@ -84,7 +84,10 @@
       headers.forEach(function(value, name) {
         this.append(name, value)
       }, this)
-
+    } else if (Array.isArray(headers)) {
+      headers.forEach(function(header) {
+        this.append(header[0], header[1]);
+      }, this);
```
mislav: We don't use semicolons.

tschaub:
Thanks for pulling this in @dgraham.  Any chance for an updated release with the fix?

dgraham:
The 2.0.3 release is now available on npm.

Issue #489, opened 2017-03-02
fails to parse headers containing line feed


pseidemann:
hi,
I have some quite old user agent `Mozilla/5.0 (QtEmbedded; Linux) AppleWebKit/534.34 (KHTML, like Gecko) webkit/0.1 Safari/534.34` (some proprietary set-top box) which seem to include line feed between cookies when the server sends multiple `Set-Cookie` headers. I know that the new spec is that the user agent should join these headers with a comma `,` but it seems like before this was `\n`.
line feeds are deprecated but still allowed per https://tools.ietf.org/html/rfc7230.
if I change the regex for splitting the headers returned by `xhr.getAllResponseHeaders()` from `/\r?\n/` to `/\r\n/` the problem is solved for this user agent https://github.com/github/fetch/blob/eebaa2a1bc21eeba98ee00c9f94a0a4c2007cff1/fetch.js#L361.
from the specs it seems headers should always be delimited by `\r\n` so I wonder if this is the proper fix for my user agent.
without the fix, my requests to the server fail while using `credentials: 'same-origin'` because cookies are not send like they were received from the server.

mislav:
We did avoid splitting by `\n` before https://github.com/github/fetch/pull/449 /cc @woudini. Now you're suggesting to revert that change. From my perspective, I can see both sides of the argument, but I'm not sure which approach is the more "correct" one.

However, I'm really surprised that you can read cookies from the server response (regardless of the value having line breaks or not). The browser usually doesn't let you read the `Set-Cookie` response header or set the `Cookie` request header because it's listed among "forbidden" headers for XHR. In other words, you are not supposed to interface with cookies directly, since that's the responsibility of the browser.

Could it be that your user agent is so old that it doesn't implement the concept of "forbidden" headers?

pseidemann:
@mislav yes I think the browser is too old to know about forbidden headers.
what came to my mind is maybe a way which would allow both ways. what if we check first if the headers string contains any `'\r'` at all, and if so we split at `'\r\n'` (response follows spec strictly), otherwise we split on `/\r?\n/` as before (tolerant version as per https://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html#sec19.3).
would this be an option?

mislav:
[...]

PR #491, opened 2017-03-06
Support obs-fold as header delimiter

woudini:
For issue https://github.com/github/fetch/issues/489


mislav:
This looks good; thank you!

pseidemann:
@woudini thanks for the fix!
I might be wrong but I think the regex missing the `g` option:
`/\r?\n[\t ]+/` -> `/\r?\n[\t ]+/g`

PR #502, opened 2017-04-05
Allow `:` in headers

XadillaX:
Here's a request's header:

```
:authority:jandan.net
:method:GET
:path:/ooxx
:scheme:https
accept:text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8

...
```

see https://jandan.net/ooxx

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -46,7 +46,7 @@
     if (typeof name !== 'string') {
       name = String(name)
     }
-    if (/[^a-z0-9\-#$%&'*+.\^_`|~]/i.test(name)) {
+    if (/[^a-z0-9\-#$%&'*+.\^_`|~\:]/i.test(name)) {
       throw new TypeError('Invalid character in header field name')
     }
     return name.toLowerCase()


XadillaX:
anyone to review this code?

mislav:
The `:` character should not be allowed in header names. Why do you need this?

XadillaX:
Imagine I am a crawler. And open this page in Chrome:

https://jandan.net/ooxx

Open developer tools and see the Network.

image

I should simulate it but failed with `fetch`.

XadillaX:
How's the result?

dgraham:
This [StackOverflow post](http://stackoverflow.com/questions/19028068/illegal-characters-in-http-headers#19028585) and linked RFCs define a header `field-name` as a `token`, which is any character except control characters and separators, including the colon. So I don't think we should add this separator character to the regular expression.

XadillaX:
So as a crawler, how could I simulate the request to https://jandan.net/ooxx by using **fetch**? Because that website need this header.

Issue #503, opened 2017-04-06
Headers are normalized to lower case against Fetch spec


piotrbazan:
Fetch spec defines that searching header in list should be case insensitive
> A header list (list) contains a name (name) if list contains a header whose name is a byte-case-insensitive match for name.

Still appending/setting header should be as is without modification:
> Append a new header whose name is name and value is value to list.

Currently all headers are normalized before adding/setting to lower case e.i.
'Content-Type' becomes 'content-type'.

which is against spec.

mislav:
It's not clear to me whether this actually "breaks" the spec. The spec doesn't dicatate that headers be sent with the original case. It indicates that _searching_ within headers is case-insensitive to conform with the HTTP spec which dictates that header names are **generally** case-insensitive, i.e. it doesn't matter which case you set the header name with (the client doesn't care about case-sensitivity, the protocol doesn't care, the server doesn't care).

In fact, if you try to set a `X-CuStOm: foo` header with Chrome's implementation of `fetch`, you will notice that the header gets sent as `x-custom: foo`. If Chrome can afford to "normalize" the casing internally (probably to simplify their implementation), I feel that so can this polyfill.

I'm closing this due to not convinced that anything is actually broken by this polyfill's behavior regarding header names. If some server requires header names in specific case, that server's implementation of HTTP is broken.

If you can argue that there is a technical reason why we should preserve original case, please comment here, and I will reopen.

PR #519, opened 2017-05-15
Remove toLowerCase of header name

ds8k:
Fix for issue #503


mislav:
Sorry, but this breaks the build, and doesn't actually fix anything. See original issue for discussion.

PR #521, opened 2017-05-17
Add additional fallbacks when self is undefined

tqc:
This extends the solution in #253 for React Native compatibility. 

The existing `this` fallback is inadequate because in strict mode `this` is undefined outside of class methods rather than defaulting to the global object.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,6 +1,10 @@
 (function(self) {
   'use strict';
 
+  if (!self) {
+    throw new Error('Fetch polyfill was unable to access the global object');
+  }
+
   if (self.fetch) {
     return
   }
@@ -461,4 +465,9 @@
     })
   }
   self.fetch.polyfill = true
-})(typeof self !== 'undefined' ? self : this);
+})(
+/* jshint ignore:start */
+/* we expect some of these to be undefined */
+  typeof self !== 'undefined' ? self : typeof this !== 'undefined' ? this : typeof global !== 'undefined' ? global : typeof window !== 'undefined' ? window : null
+/* jshint ignore:end */
+);


mislav:
I have no experience with React Native, but their documentation suggests that `fetch` is available natively. If this is true, why is the polyfill needed?

tqc:
@mislav The polyfill itself may not be needed, but libraries that support multiple platforms (including react itself) can't know that, and without self the polyfill crashes at the point where you check if it is needed.

tqc:
@mislav incidentally, I don't think that documentation implies fetch is available natively - just that react wraps this polyfill so the app developer doesnt have to think about it.

psullivan6:
**Related to this conversation:**
Using this polyfill with rollup logs the following warning during the build:

```
node_modules/whatwg-fetch/fetch.js (461:40) The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten
```

which means the final line of the script is compiled into:  
`})(typeof self !== 'undefined' ? self : undefined);` where `undefined` replaced `this`.

---

This is using the ES6 `import 'whatwg-fetch';` statement.

AntonioRedondo:
Any idea when is this gonna be merged? We're trying to use `whatwg-fetch` server-side bundled with Rollup. We're not sure we'll be finally able to use it. But this issue is definitely a blocker for our try.

mislav:
We now generate a UMD build with Rollup: https://github.com/github/fetch/pull/616. Let us know if this build works with React Native.

The main [`fetch.js` file](https://github.com/github/fetch/blob/master/fetch.js) is an ES5 file (like before) but with added ES6 exports. This makes it compatible with Rollup, and is referred to by `module` keyword from `package.json`. This should make the library work better with packagers.

PR #526, opened 2017-05-24
Ensure cookies aren't sent if `credentials: omit`

satya164:
If the default value of `withCredentials` is `true`, as opposed to `false`, the library will send the cookies even if `credentials: omit` is specified. This affects React Native where the default value is `true` since cross-origin concerns don't apply.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -447,6 +447,8 @@
 
       if (request.credentials === 'include') {
         xhr.withCredentials = true
+      } else if (request.credentials === 'omit') {
+        xhr.withCredentials = false
       }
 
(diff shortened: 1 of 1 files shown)


ericvicenti:
Yes, we very much need this for React Native's `fetch` because the `XMLHttpRequest.sendRequest` default value of `withCredentials` is `true` to match the native platforms' default behavior.

Thank you @satya164!

mislav:
Makes sense; thanks!

misogl:
Just ran into this issue in React Native. Could you please publish a new version of this package with this fix?
Thanks.

PR #534, opened 2017-06-16
Allow undefined Response status

remcohaszing:
This allows the Response constructor to accept a status code that's
explicitly set to undefined. Native fetch implementations in Chrome and
Firefox behave this way.

The following now works, but would previously throw an error:

    new Response('', { status: undefined })

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -380,7 +380,7 @@
     }
 
     this.type = 'default'
-    this.status = 'status' in options ? options.status : 200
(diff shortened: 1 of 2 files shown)


mislav:
Thank you!

Nader-Zubaid:
#535

Nader-Zubaid:
#_**# 534**_

wasem1772:
@github-cla

PR #543, opened 2017-07-20
Add link to cdnjs to README

kribblo:
For when you just want to throw it quickly into a page, want a quick download or don't use any of those extra tools, the cdnjs service is also available

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,9 @@ expected to uphold this code.
 
 * `bower install fetch`; or
 
-* `yarn add whatwg-fetch`.
+* `yarn add whatwg-fetch`; or
+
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the suggestion, but we won't be adding cdnjs instructions right now.

We recommend installation approaches that are directly under our control, e.g. npm package. Cdnjs is a service that is outside of our control, i.e. it is not guaranteed that it will stay up for the foreseeable future nor that they will keep the version of fetch up to date.

If someone is already familiar with cdnjs, it's not hard for them to figure out how to load fetch over their service. For everyone else, installing and loading fetch from within their app is still the approach I would personally recommend for the long term.

Issue #180, opened 2015-07-19
Export fetch function


jpochyla:
Exporting the `fetch` function through `module.exports` would allow me to use the polyfill instead of buggy browser implementations. For example, current Chrome Canary incorrectly throws on 504 response, without any sensible way of detecting the cause of error in the catch handler.

dgraham:
Please open a bug report with the Chrome team regarding the incorrect behavior. This polyfill is not intended to mask browser bugs, only to provide the `window.fetch` global in browsers without any `fetch` implementation at all.

jpochyla:
Can I ask what is the reasoning behind that? Hiding the implementation behind a global and making it absolutely impossible to access otherwise seems overly limiting, and not adding anything.

runn1ng:
If I understand @jpochyla correctly, he is mostly complaining that the `fetch` cannot be accessed on browsers that already have `fetch`. There is no reason this module cannot export itself via `module.exports`.

mislav:
@jpochyla Here is how to force the polyfill.

```html
<script>window.fetch = undefined</script>
<script src=fetch.js></script>
```

@runn1ng We don't feel that a module pattern is worth maintaining for this simple polyfill. Any faulty browser implementation of `fetch` is that browser vendor's problem and should be taken with them.

runn1ng:
@mislav well that's a nice hack :)

bamsebjrn:
excitera me sir but what? im jusing tor o chrome

Skickat från min iPhone

8 aug. 2018 kl. 09:19 skrev Jarek Rencz <notifications@github.com<mailto:notifications@github.com>>:


@mislav<https://github.com/mislav> thanks for a hint :)

Any faulty browser implementation of fetch is that browser vendor's problem and should be taken with them.

To a certain extent it's not always on the browser vendor's side. E.g.: In a project of mine, where I use fetchI experience a strange behaviour in Firefox <54 where using native fetch it produces POST requests without payload. I was able to reproduce it with 1/500-1000 ratio but I see it in production so it's a fact.

Enforcing polyfill allows me to make my app usable for such users (those are mostly Firefox 52 users)

And now to the point: Firefox 52 is the last one on Win XP and Win Vista. I don't have a comfort of telling my users to update their PCs' OS just because at some point Mozilla did something bad in fetch implementation and then they fixed it (to my best knowledge either silently or accidentally because I found no bug report about such behaviour in Bugzilla)

mislav:
[...]

PR #545, opened 2017-07-20
add UMD on package

qraynaud:
I think this is a good alternative for both #521 and #180.

Adding and UMD allows the polyfill to be loaded in more environments and let the user choose between their native implementation and the polyfill if they want and whenever they want.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -20,6 +20,9 @@
   "worker": true,
   "globals": {
     "JSON": false,
-    "URLSearchParams": false
+    "URLSearchParams": false,
+    "define": true,
+    "module": true,
+    "global": true
   }
 }
--- a/fetch.js
+++ b/fetch.js
@@ -1,14 +1,27 @@
-(function(self) {
+(function (root, factory) {
   'use strict';
 
-  if (self.fetch) {
-    return
+  if (typeof define === 'function' && define.amd) {
+    // AMD. Register as an anonymous module.
+    define([], function() {
+      return factory(root)
+    })
+  } else if (typeof module === 'object' && module.exports) {
+    // Node. Does not work with strict CommonJS, but
+    // only CommonJS-like environments that support module.exports,
+    // like Node.
+    module.exports = factory(root)
+  } else {
+    // Browser globals (root is window)
+    root.fetchPolyfill = factory(root)
   }
+}(typeof self !== 'undefined' ? self : this, function (root) {
+  'use strict';
 
   var support = {
-    searchParams: 'URLSearchParams' in self,
-    iterable: 'Symbol' in self && 'iterator' in Symbol,
-    blob: 'FileReader' in self && 'Blob' in self && (function() {
+    searchParams: 'URLSearchParams' in root,
+    iterable: 'Symbol' in root && 'iterator' in Symbol,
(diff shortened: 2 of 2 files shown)


qraynaud:
I don't really understand why some of those tests fails. I don't think I'm responsible for this.

qraynaud:
Any input on this?

fetch.js
```diff
@@ -463,4 +489,6 @@
     })
   }
   self.fetch.polyfill = true
-})(typeof self !== 'undefined' ? self : this);
```
mislav: You've removed this construct which was here to make `fetch` work for both normal browser and [webworker](https://github.com/github/fetch/blob/27473e937c6bdbb61212cc078ad60d22ae77ffb1/test/test-worker.html) environments. You've replaced it with two different contexts: `this` for the concept of the "root" in the UMD shim, and `window
 || global` for the concept of "main" inside the factory function. This is likely what broke the test suite for webworker.

I think the concepts of "root" and "main" should be unified and should be passed around as a single reference. Also, per https://github.com/github/fetch/pull/521, I think the fallback chain for getting a reference to the global object should be (pseudocode) `self || this || global`.

.jshintrc
```diff
@@ -20,6 +20,9 @@
   "worker": true,
   "globals": {
     "JSON": false,
-    "URLSearchParams": false
+    "URLSearchParams": false,
+    "define": true,
+    "module": true,
+    "global": true
```
mislav: Isn't there a jshint "environment" declaration akin to `"worker": true` that makes it assume that these globals are available; possibly `"umd": true` or `"amd": true`?

fetch.js
```diff
@@ -1,23 +1,47 @@
...
+  var main = {};
+
+  try {
+    main = window
+  } catch(e) {
+    main = global
   }
 
   var support = {
-    searchParams: 'URLSearchParams' in self,
```
mislav: This unnecessarily increases the size of the diff. Try naming the global object `self` instead of `main`. Also, please pass the global object by argument into the factory function instead of determining it inside the factory.

.jshintrc
qraynaud: Sadly, I didn't find any. There are a lot of things like that in eslint but I cannot find those for jshint.

fetch.js
qraynaud: I cannot name it self because I already defined it later to avoid diffs on more important stuff (since what was self is now split between `self` = what is exported and `main` = global object). I'll rename it `root` while doing the other changes you suggest so it keeps the same nomenclature as the UMD definition.

fetch.js
qraynaud: Done

qraynaud:
Seems like now all tests are green ! Thanks.

I also moved the code that exported in global the polyfill to the main codebase so it happens for AMD & commonjs too just in case some people already loaded the file in such environments one way or another and expected it to auto export fetch.

mislav:
@qraynaud Thank you for your contribution, and sorry that this didn't get merged. We weren't ready to add a UMD wrapper back then, but we were ready now, since a need arose to have module exports (specifically: abortable fetch requests). However, we've made a build step that generates a UMD build via Rollup. https://github.com/github/fetch/pull/616

PR #550, opened 2017-08-01
add normalization for PATCH method

dfguo:
Previously when passing lowercase method `patch`, the library will let it pass as it is. This causes some reverse proxy to reject the request. 

This change expanded the normalization function to include `patch`

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -295,7 +295,7 @@
   }
 
   // HTTP methods whose capitalization should be normalized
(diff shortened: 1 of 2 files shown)


mislav:
We only normalize method names which [the spec dictates to normalize](https://fetch.spec.whatwg.org/#concept-method-normalize):

>  if it is a byte-case-insensitive match for `DELETE`, `GET`, `HEAD`, `OPTIONS`, `POST`, or `PUT`, byte-uppercase it.

Your proposed change would deviate from the spec. Instead, be sure to always use uppercase `PATCH` when calling `fetch()`.

PR #551, opened 2017-08-01
Add caveat about disabling same-origin cookies

kdama:
Because of the spec of XMLHttpRequest, `fetch` polyfill can't prevent sending and receiving same-origin cookies when `credentials: 'omit'`.
This behavior is very confusing when app supports both modern and legacy browsers.
I think this should be clearly documented as a caveat.


mislav:
Thank you!

PR #558, opened 2017-09-06
WebWorker jsonParse

lifeart:
Non-blocking JSON parsing - https://gist.github.com/lifeart/716d7f0bb087c34247487e4abee2419c

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -4,7 +4,7 @@
   if (self.fetch) {
     return
   }
-
+  var hasWebWorker = typeof window.Worker !== 'function';
   var support = {
     searchParams: 'URLSearchParams' in self,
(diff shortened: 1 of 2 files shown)


mislav:
Thank you for the suggestion!

This is a very neat trick, but I think it's out of the scope for our polyfill, which we're trying to make small and simple. JSON parsing will remain to be “blocking” because JSON parsing usually isn't the bottleneck for most applications out there. For applications that _do_ have JSON parsing as the bottleneck, they can choose to not invoke `response.json()` and instead invoke `response.text()` and perform JSON parsing themselves. So you see, we believe that such optimizations should be performed only when absolutely needed, and not in the polyfill itself.

PR #562, opened 2017-09-14
Change error message for xhr.ontimeout

apoh:
At the moment xhr.onerror and xhr.ontimeout do write the same error
message of TypeError('Network request failed') to distiguish between
this commit changes the error message for the timeout.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -440,7 +440,7 @@
       }
 
       xhr.ontimeout = function() {
-        reject(new TypeError('Network request failed'))
+        reject(new TypeError('Network request timed out'))
       }
 
       xhr.open(request.method, request.url, true)


mislav:
Thank you for your contribution. I don't, however, think that this useful.

First of all, the `ontimeout` handler is only effective in Safari: 619ee08d5102f3a1540dd515a1fec766f3c01737.

Second, improving this exception message in our polyfill will only create an illusion that timeouts and other request errors can be distinguished between each other, but they can't. In native implementations of `fetch()` in modern browsers, Firefox has an identical message “NetworkError when attempting to fetch resource” for all kinds of connection errors, including timeouts; Safari 10.1 shows the “Type error” message in all these conditions; Chrome simply shows “Failed to fetch”.

In none of these browsers is it possible to distinguish between timeouts and other kinds of errors. Therefore, this polyfill will not attempt to distinguish either.

dewwwald:
Follow my thinking regarding this issue on this thread over here: #565 , I think that issue is much more coherent and providers a more general issue to be solved.

Issue #539, opened 2017-07-05
Can't understand Readme installation part


gkatsanos:
You mention

> For use with webpack, add this package in the entry configuration option before your application entry point:

entry: ['whatwg-fetch', ...]

I don't understand if this is optional or complements the next part which is importing.

Also, why is it necessary to change the webpack configuration to load this specific module?

mislav:
(Not a webpack user myself) If you use webpack for your application, then you have to somehow import the fetch polyfill, right? How do you otherwise do it for other polyfills?

gkatsanos:
I thought I just do `npm install <packagename>` and then (provided the package is CommonJS/Harmony compliant) I just do `import packagename from 'packagename'`. Then webpack would look it up into node_modules and that's all..

mislav:
The package is not CommonJS/Harmony compliant. It's a polyfill, meaning that instead of exporting some functions, it writes to `window.fetch` only if `window.fetch` was missing.

Anyway, if you catch something wrong with our webpack instructions, and know some other instructions that work better than ones we've provided, please open a pull request with corrections. Otherwise, we're just speculating in this thread and not really going anywhere. Thank you!

gkatsanos:
Thanks, I ended up using Axios instead which didn't require changing
webpack configuration.

PR #569, opened 2017-10-11
Update README.md to mention global variable

turquoisemelon:
Proposing updating the installation section of the docs to mention polyfill global. There was an issue opened about this in the past: https://github.com/github/fetch/issues/539 . 
Please let me know what you think of my correction. I'd be more than happy to update it based on project maintainers' feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,18 +68,27 @@ You will also need a Promise polyfill for [older browsers](http://caniuse.com/#f
 We recommend [taylorhakes/promise-polyfill](https://github.com/taylorhakes/promise-polyfill)
 for its small size and Promises/A+ compatibility.
 
-For use with webpack, add this package in the `entry` configuration option
+For use with webpack, either add this package in the `entry` configuration option
 before your application entry point:
 
 ```javascript
 entry: ['whatwg-fetch', ...]
 ```
 
-For Babel and ES2015+, make sure to import the file:
+... or require the file:
+
+```javascript
+require('whatwg-fetch')
+```
+
+For Babel and ES2015+, import the file:
 
 ```javascript
 import 'whatwg-fetch'
 ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,19 +68,28 @@ You will also need a Promise polyfill for [older browsers](http://caniuse.com/#f
...
+```
+
+For Babel and ES2015+, import the file:
 
 ```javascript
 import 'whatwg-fetch'
 ```
 
+In either case the polyfill sets up a global variable. The module does not export
+anything directly, so do not use the `import fetch from ...` form of the import statement.
```
mislav: Is this necessary to point out to clarify webpack instructions?

README.md
```diff
@@ -68,19 +68,28 @@ You will also need a Promise polyfill for [older browsers](http://caniuse.com/#f
...
 
 ```javascript
 entry: ['whatwg-fetch', ...]
 ```
 
-For Babel and ES2015+, make sure to import the file:
+... or require the file:
+
+```javascript
+require('whatwg-fetch')
```
mislav: Is `require()` something specific to webpack? I'm not a webpack user.

turquoisemelon: It's not specific to webpack but it's part of the common usage. I pointed it out as it is another way to load the package, in addition to the first one that was included in the docs.

README.md
turquoisemelon: I thought it's a good idea to add an explanation as to why both approaches work. Especially after seeing the comments in this closed issue: https://github.com/github/fetch/issues/539

mislav: Sorry that this never got merged. The library went to some changes since then, and now it exports the `fetch()` method and associated classes. I will update the README documentation to reflect that.

PR #572, opened 2017-10-17
Added AbortController polyfill

### LINKED ISSUE #547: Implement abort
As per https://github.com/whatwg/fetch/pull/523

I guess it should wait till the above is merged.

coldhamix:
Closes #547. Used polyfill from https://github.com/mo/abortcontroller-polyfill/. Tests included.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -23,6 +23,7 @@
     "URLSearchParams": false,
     "AbortController": false,
     "AbortSignal": false,
-    "DOMException": false
+    "DOMException": false,
+    "Event": false
   }
 }
--- a/.jshintrc
+++ b/.jshintrc
@@ -20,6 +20,9 @@
   "worker": true,
   "globals": {
     "JSON": false,
-    "URLSearchParams": false
+    "URLSearchParams": false,
+    "AbortController": false,
+    "AbortSignal": false,
+    "DOMException": false
   }
 }
--- a/fetch.js
+++ b/fetch.js
@@ -472,15 +472,41 @@
   }
 
   function Emitter() {
-    var delegate = document.createDocumentFragment()
-    var methods = ['addEventListener', 'dispatchEvent', 'removeEventListener']
-    methods.forEach(function(method) {
-      this[method] = function () {
-        delegate[method].apply(delegate, arguments);
-      };
-    }.bind(this))
+    this.listeners = {}
   }
 
+  Emitter.prototype.listeners = null
+  Emitter.prototype.addEventListener = function(type, callback) {
+    if (!(type in this.listeners)) {
+      this.listeners[type] = []
+    }
+    this.listeners[type].push(callback)
+  }
+
+  Emitter.prototype.removeEventListener = function(type, callback) {
+    if (!(type in this.listeners)) {
+      return
+    }
+    var stack = this.listeners[type]
+    for (var i = 0, l = stack.length; i < l; i++) {
+      if (stack[i] === callback){
+        stack.splice(i, 1)
+        return
+      }
+    }
+  }
+
+  Emitter.prototype.dispatchEvent = function(event) {
+    if (!(event.type in this.listeners)) {
+      return true
+    }
+    var stack = this.listeners[event.type]
+
+    for (var i = 0, l = stack.length; i < l; i++) {
+      stack[i].call(this, event)
+    }
+    return !event.defaultPrevented
+  }
 
   function AbortSignal() {
     Emitter.call(this)
--- a/fetch.js
+++ b/fetch.js
@@ -464,3 +464,93 @@
   }
   self.fetch.polyfill = true
 })(typeof self !== 'undefined' ? self : this);
+(function(self) {
+  'use strict';
+
+  if (self.AbortController) {
+    return
+  }
+
+  function Emitter() {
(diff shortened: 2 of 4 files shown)


mislav:
Thanks for the suggestion! This looks like it simply pastes the other polyfill after ours. Could we instead keep the projects separate and instruct the users to compose the polyfills if they need abort functionality?

coldhamix:
I was concerned about the script order (which might be not really easy to maintain) as the fetch polyfill should always be before AbortController polyfill. Moreover abort controller is a part of fetch standard and it does not feel really good to me to separate its functionality into several polyfills.

mislav:
Sure, there is some convenience having the complete `fetch` polyfill in one place. But there are also benefits of keeping them separate, such as better maintainability over time, and loading order shouldn't be a problem for anyone who loads JS files in their app.

@mo @jimmywarting What do you think about this?

web-devel:
@mislav don't you think that once abort signal is landed in browsers it will be strange not to see it in this polyfill

jimmywarting:
I think that the abortable controller shouldn't be part of this fetch polyfill
A abortable controller dosen't have anything to do with fetch, it's meant to be used with many other promise based api's

However I think this fetch polyfill should be able to handle a abort if a person send a AbortSignal in as an options and actually call `xhr.abort()` when a signal is aborted

The problem with this PR is that while it rejects and dose what it is suppose to do... It actually doesn't close the connection when the request is aborted (as mention in the readme.md)

It's up to every promise based api to handle a AbortSignal.

--- 

And when it comes to order of execution... AbortController should load before fetch.

jimmywarting:
This is what I think should be implemented

```js
function abortError(){
  var err;
  try {
    err = new DOMException('Aborted', 'AbortError');
  } catch (err) {
    // IE 11 does not support calling the DOMException constructor, use a
    // regular error object on it instead.
    err = new Error('Aborted');
    err.name = 'AbortError';
  }

  return err
}

// ...

  var request = new Request(input, init)
  
  if (request.signal) {
    // Return early if already aborted, thus avoiding making an HTTP request
    if (request.signal.aborted) {
      return reject(abortError());
    }
    
    request.signal.addEventListener('abort', function () {
      reject(abortError())
      xhr.abort();
    }, {once: true});
  }

  // rest of the code
})
```

And also able to pass the signal into new request when cloning them if that is not taken care of...

coldhamix:
@jimmywarting @mislav How should it behave in case when there is already a native fetch implementation, but it does not support AbortController? Ignore it and polyfill with XHR?

jimmywarting:
The only way to abort a request today is by using xhr, You can't even use native fetch implementation. Firefox will be first to release it in there next version (v57)

This code [here](https://github.com/mo/abortcontroller-polyfill/blob/3906ca2f9f28009bacf1a15c918221a1bcba99cd/dist/abortcontroller.js#L84-L118) should really only be implemented by a simple test to see if Request supports signal, that way it can patch native fetch and not doing it for a polyfill like this if it's dose what it's supposed to do

eg:

```js
var controller = new AbortController();
var signal = controller.signal;
if (window.Request) {
  var request = new Request('/', { signal });

  if (!request.signal) {
    // Native implementation don't support signal
    // patch it only then
  }
}
```

just-boris:
While I very much appreciate the work is going on here, I would like to mention, that the proposed solution doesn't fully match the spec. Instead of actually aborting request, it just simulates the aborted response into the promise, but the request is being made.

Instead of appending the abortable part at the end of the file, I would suggest modifying the original fetch function. Moving change into the main part would give an access to the inner `xhr` object, so it may become truly abortable.

`AbortController` can be still shipped separately and can be included into a page in any order because there will be no direct dependencies and relying on object presence.

jimmywarting:
@just-boris Pretty much what I have been saying already

mo:
I've prepared an abortcontroller-polyfill branch based on @jimmywarting's patch that broke AbortSignal + AbortController out into a separate file so it can be imported without also getting a patched fetch():

https://github.com/mo/abortcontroller-polyfill/tree/break-out-abortcontroller/src

I will only merge that to abortcontroller-polyfill master if it's useful for you; if you go in another direction I will leave abortcontroller-polyfill master as-is.

Attrash-Islam:
The Abort feature is in the fetch spec. therefore I think that make it in another polyfill won't match the specs.

https://developers.google.com/web/updates/2017/09/abortable-fetch

jimmywarting:
@Attrash-Islam the fetch spec is only covering accepting an abort signal. Not how a abortcontroller is made.

A abort controller can be used for so much more then just the fetch api. fetch api so just happens to be the first public usecase where an abort signal can actually take places in any native javascript function.

Next time link to an actual spec and not to some blog post covering it's use cases.

jmeas:
I attempted another pass at a resolution to #547 over in #592 based on some of the feedback here.

- It doesn't require an `AbortController` or `AbortSignal` polyfill to be included in this library
- It actually aborts the request
- Minimal code changes
- Still based on the excellent polyfill over at https://github.com/mo/abortcontroller-polyfill/

I'm curious to hear what others think of it!

mislav:
Abortable fetch requests are now supported after https://github.com/github/fetch/pull/592. Note that AbortController polyfill needs to be loaded separately in order for this to work in older browsers.

PR #575, opened 2017-10-23
Fix issue #533

### LINKED ISSUE #533: In IE11 resolve is not always executed
Our app is making scheduled request every N seconds. After some time IE11 stops doing requests.
I try to search web and it looks like there is a bug in IE11. It does not handle XHR callbacks properly.
See similar issue on [bluebird repository](https://github.com/petkaantonov/bluebird/issues/1161).
So I try workaround - wrap resolve to setTimeout and it works.

CrOrc:
As stated in petkaantonov/bluebird#1161, resolving a promise in an XHR callback without active DOM operations will not trigger the promise chain to execute. But wrapping fetch promise resolution in setTimeout magically resolves it.

Experienced this issue with different promise polyfills: bluebird, es6-promise, core-js/fn/promise

Adding setTimeout fixes #533.


xemasiv:
:+1:  for this

ADagen:
Hey guys, let's move on.

wmertens:
👍 looks great, we still have to support IE11 so I don't understand why this is not merged?

ILoveITScience:
Whats the status of the issue? We also have to suport IE 11 and need this fix.

wmertens:
@mislav 	what is your opinion of this?

ILoveITScience:
I'm testing https://github.com/stefanpenner/es6-promise at the moment. It seem to fix my problem.

dasa:
This may no longer be an issue in newer versions of IE 11. I'm not able to repo using the test at: https://github.com/lowkay/MutationObserverSchedulerBug

CrOrc:
Errors in build are not caused by my changes

> HeadlessChrome 73.0.3683 (Linux 0.0.0) native Response default status is 200 OK FAILED
HeadlessChrome 73.0.3683 (Linux 0.0.0) native Response default status is 200 OK when an explicit undefined status code is passed FAILED

Default native response statusText is '' instead of 'OK' for HeadlessChrome 73.0.3683 (Linux 0.0.0).
Everything seems to work for HeadlessChrome 72.0.3626 (Windows 7.0.0).

CrOrc:
See #696 Change Response's statusText's default

superzadeh:
Any update on this? I'm also running into a bug caused by this (described in #533) and would need this.

nomnom11:
what is this problem

ntucker:
This breaks jest timer mocks

ntucker:
setTimeout() here is only called after resolution. but the promise exposed here doesn't resolve until after it is finished. Therefore, there's no way to tell jest to advance the setTimeout timer, since it isn't a timer to advance until a time to which there is no user level hookup

aldarmak (commented):
Is it the key

Issue #576, opened 2017-10-23
Response contructor is incompatible with other browser implementations


spearmootz:
if you use new Response directly and use 204 responses you get the following issue.

`new Response('', Init)`
fails on chrome because it expects no body. you can do

`new Response(null, init)`
fails on Firefox because body cannot be null

but you can do in both firefox, chrome, Edge, safari
`new Response(init);`

but that fails when using this polyfill.

spearmootz:
i get around it with the following code.

```const polyfilled = fetch.toString().includes('XMLHttpRequest');
if (polyfilled) {
  const getClassOf = Function.prototype.call.bind(Object.prototype.toString);
  const isClassOf = (instance, className) => getClassOf(instance).split(' ')[1].includes(className);
  const PolyfilledResponse = window.Response;
  window.Response = (first, second) => {
    if (isClassOf(first, 'Object')) {
      return new PolyfilledResponse(null, first);
    }

    return new PolyfilledResponse(first, second);
  }
}```

mislav:
Thanks for reporting! In the meantime, I think you can use `new Response(undefined, {status: 204})` as a workaround in your app. It's a shame null doesn't work in Firefox; is this known to Mozilla?

spearmootz:
so my original assumption that doing new Response(options) works on other browsers. it actually returns a response that is a 200, no matter if you did 204 or 400. so it will always be resp.ok

dgraham:
Reopening to track updating our `Response` constructor to match the [specification changes](https://fetch.spec.whatwg.org/#response-class).

We need to handle the optional first `BodyInit` argument here.

https://github.com/github/fetch/blob/7a692dc519a3a110526a26c7e252f99991510c2a/fetch.js#L377-L380

PR #577, opened 2017-10-23
Allows for first new Response argument to be options

spearmootz:
Fixes https://github.com/github/fetch/issues/576

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -42,12 +42,11 @@
     }
   }
 
-    var getClassOf = Function.prototype.call.bind(Object.prototype.toString)
+  var getClassOf = Function.prototype.call.bind(Object.prototype.toString)
 
-    var isClassOf = function (instance, className) {
-      return getClassOf(instance).split(' ')[1] === className + ']'
-    }
-
+  var isClassOf = function (instance, className) {
+    return getClassOf(instance).split(' ')[1] === className + ']'
+  }
 
   function normalizeName(name) {
     if (typeof name !== 'string') {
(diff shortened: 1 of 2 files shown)


fetch.js
```diff
@@ -42,6 +42,12 @@
     }
   }
 
+  var getClassOf = Function.prototype.call.bind(Object.prototype.toString)
+
+  var isClassOf = function (instance, className) {
+    return getClassOf(instance).split(' ')[1] === className + ']'
```
mislav: Could we get away with a simpler check?

```js
function isPlainObject(obj) {
  return obj != null && typeof obj === 'object' && Object.prototype.toString.call(obj) === '[object Object]'
}
```

fetch.js
```diff
@@ -374,7 +380,19 @@
 
   Body.call(Request.prototype)
 
-  function Response(bodyInit, options) {
+  function Response() {
+    var bodyInit
+    var options
+
+    // if the first argument is options and not body
+    if (isClassOf(arguments[0], 'Object')) {
```
mislav: How about something like

```js
function Response(bodyInit, options) {
  if (arguments.length === 1 && isPlainObject(bodyInit)) {
    options = bodyInit
    bodyInit = null
  }
  // ...
}
```

test/test.js
```diff
@@ -603,6 +603,12 @@ suite('Response', function() {
 
     assert.equal(r.headers.get('content-type'), 'text/plain')
   })
+
+  test('accepts options as first parameter', function () {
+    var r = new Response({ status: 200 })
```
mislav: Let's use a non-default status such as 204 to verify that the status was successfully assigned.

PR #592, opened 2017-12-07
Support abort API

### LINKED ISSUE #547: Implement abort

jamesplease:
This is a quick pass at an attempt to resolve #547 . The intention is that it would supersede #572

Todo:

- [ ] Tests
- [ ] ~~Verify that the feature detection works~~
- [x] Update documentation

Feedback is welcome! Thanks for reviewing :v:


fetch.js
```diff
@@ -1,7 +1,21 @@
 (function(self) {
   'use strict';
 
-  if (self.fetch) {
+  function canAbortFetch() {
+    if (!self.AbortController || !self.AbortSignal) {
```
jmeas: I suppose we may also need to check for `self.Request` here as well?

fetch.js
```diff
@@ -1,7 +1,21 @@
...
+      return false
+    }
+  
+    var abortController = new self.AbortController()
+  
+    var request = new self.Request('http://a', {
```
jmeas: Presumably, the truthiness of `self.fetch` from L18 indicates that `self.Request` exists. To be more explicit, we could add `self.Request` to L5 if that's preferable.

fetch.js
```diff
@@ -1,7 +1,21 @@
...
+      signal: abortController.signal
+    })
+  
+    return Boolean(request.signal)
+  }
+
+  if (self.fetch && canAbortFetch()) {
```
mislav: Does this entirely overwrite the native `window.fetch` implementation if fetch isn't abortable?

I don't think that's acceptable. This polyfill should perhaps *wrap* native fetch to provide it with abort functionality, but not overwrite it entirely.

jmeas:
Thanks for the feedback, @mislav . Sure, I can update this PR. I’ll go for item 1 that you listed.

One concern I have is that if we wrap a native fetch, we will not be able to _actually_ cancel the request, which is an important feature, I think.

fetch.js
jmeas: Correct, that is the current behavior. The intention here was to ensure that the HTTP requests are _actually_ canceled to reduce the amount of data being sent over the network.

fetch.js
```diff
@@ -459,6 +463,10 @@
         xhr.setRequestHeader(name, value)
       })
 
+      if (init.signal && init.signal.addEventListener) {
+        init.signal.addEventListener('abort', xhr.abort)
```
jimmywarting: Should we remove the event listener also in the `xhr.onabort` function?

jimmywarting:
This PR looks good to me 👍 
I would merge this.

fetch.js
jmeas: Yup. I'll do that.

jimmywarting:
Just thought about another thing. 

You are using `init.signal` instead of `request.signal`

Fetch can also be called as followed:

```js
var req = new Request(url, {signal})
fetch(req)
```

Guessing you have to dig down to the Request constructor and set `this.signal = options.signal`

jmeas:
Good catch @jimmywarting ! I'll add that.

jimmywarting:
One thing that baffle me was that Firefox (that supports AbortController) sets a default abortSignal to the Request even if no signal was passed in.

`new Request('http://a').signal // a signal` 😕 

I don't know if this is intentional by spec. haven't looked at the spec lately
I also expected `Request.prototype.signal` to be `null`. Same goes for new Request('http://a').signal == null // false

jmeas:
> I don't know if this is intentional by spec.

This is intentional. From [this page](https://developers.google.com/web/updates/2017/09/abortable-fetch):

image

This is great, because it makes [the feature detection code](https://gist.github.com/jmeas/39546faa735779051bd00c3fe2a136f9) so small!

jmeas:
@jimmywarting , this has been updated with your suggestions. I think this covers the two cases you brought up:

1. Passing in a `Request` as the first argument
2. Bailing early if the Controller was aborted before calling `fetch`

jmeas:
I added a section to the documentation about this. I wasn't too sure where to put it, so I threw it into the caveats. Happy to move that around or reword it or whatever.

When it comes to testing, what would folks like to see here?

fetch.js
```diff
@@ -459,6 +469,10 @@
         xhr.setRequestHeader(name, value)
       })
 
+      if (request.signal && request.signal.addEventListener) {
+        request.signal.addEventListener('abort', xhr.abort)
```
mislav: Here, `event` object will be passed to `xhr.abort()` as an argument, which probably doesn't have side-effects, but to avoid it let's create a separate function for the event handler that calls into `xhr.abort()`.

fetch.js
```diff
@@ -424,6 +425,10 @@
       var request = new Request(input, init)
       var xhr = new XMLHttpRequest()
 
+      if (request.signal && request.signal.aborted) {
+        return reject(new DOMException('Aborted', 'AbortError'));
```
mislav: Is this style of DOMException interface supported in all browsers that fetch aims to support?

fetch.js
```diff
@@ -459,6 +469,10 @@
         xhr.setRequestHeader(name, value)
       })
 
+      if (request.signal && request.signal.addEventListener) {
```
mislav: Can we assume that if `request.signal` isn't null, it will always define `addEventListener`?

README.md
```diff
@@ -260,6 +261,47 @@ response.headers['X-Request-URL'] = request.url
...
+const controller = new AbortController()
+
+fetch('/avatars', {
+  method: 'POST',
+  body: data,
+  signal: controller.signal
+})
+  .then(
+    function(response) {
+      console.log('request succeeded', response)
```
mislav: Could we shorten the `fetch` invocation example to a minimum possible code that checks for aborted requests?

```js
fetch('/avatars', {
  signal: controller.signal
}).catch(function(error) {
  if (error.name === 'AbortError') {
    // request was aborted
  }
})
```

mislav:
> When it comes to testing, what would folks like to see here?

Please:
1. Create a “slow” route in `script/server`, make a request, abort it before it hand a chance to complete;
2. Create a Request with already aborted controller, verify that it's immediately rejected.

fetch.js
jmeas: I can do some investigating this weekend and get back to you on that.

README.md
jmeas: Of course!

fetch.js
jmeas: Sure thing :v:

jmeas:
Latest changes:

- `xhr.abort` is now called without any arguments
- simplified example code for aborting fetches
- assumptions are made that `request.signal` implements the Event interface

To do:

- Add some tests according to @mislav's specifications above
- Investigate that all target browsers support `DOMException(message, name)`. The big ? here is the second argument support. IE9 does not support it afaik, so it'll be interesting to see if 10 does!

@mislav, assuming IE10 does not support that API of DOMException, what would you propose as the workaround?

jmeas:
@mislav, when it comes to the AbortController/Signal polyfill, what would you like me to do there?

I could write a quick polyfill here since the API is so small, or I could pull in the one from [this library](https://github.com/mo/abortcontroller-polyfill), or any other library. Let me know and I'll get to writing those tests!

mislav:
> when it comes to the AbortController/Signal polyfill, what would you like me to do there?

I still haven't decided, but for this PR, let's skip bundling the AbortController polyfill, and decide that in subsequent PRs.

> IE10 does not support DOMException, so we need a different solution there.

Yes, and PhantomJS seems to fail for the same reason in CI. Your proposed solution looks good, but why the `if (!ex.name)`? Isn't setting the `name` important enough that it shouldn't be a conditional?

jmeas:
> I still haven't decided, but for this PR, let's skip bundling the AbortController polyfill, and decide that in subsequent PRs.

Sorry, to clarify, I meant for the tests, rather than in the fetch code itself.

> Yes, and PhantomJS seems to fail for the same reason in CI. Your proposed solution looks good, but why the if (!ex.name)? Isn't setting the name important enough that it shouldn't be a conditional?

Whoops. That's left over from an earlier iteration of the code, when I thought it might support `DOMException`, but not the `name` argument. The `(!ex.name)` won't be there :v:

mislav:
I've pushed some commits to illustrate how I would go about this. Let me know what you think

jmeas:
Thanks @mislav ! That's super useful. I've been busy with the holidays, but I'll take a look at this again shortly.

fonzy2013:
Is this planned to be released soon? Is it using the AbortController and AbortSignal interfaces?

jmeas:
> Is this planned to be released soon?

@fonzy2013 , first, the PR needs to be completed and merged. There's still a bit of work to do here.

> Is it using the AbortController and AbortSignal interfaces?

You can read the source code or the comments in this issue/linked issues to determine that yes, it is 😉 

---

I've been a bit busy; haven't had time to swing back around to this. It's still on my todo list, but I'll likely be preoccupied for a few more weeks. If someone else wants to pick it up in the meantime, then feel free!

joaovieira:
@jmeas @mislav what is missing? how can we help?

jmeas:
@joaovieira there is an up-to-date todo list in the opening comment of this PR. It needs some more tests to verify all of the new function. @mislav put some example tests in that can be used as a start to add the rest.

joaovieira:
Added a fix up in https://github.com/jmeas/fetch/pull/1. 

Was also trying to include the [abortcontroller-polyfill](https://github.com/mo/abortcontroller-polyfill/) for testing (instead of creating dummy signals) just like the promise polyfill is included. Works just fine for the regular tests but the workers tests fail because of https://github.com/mo/abortcontroller-polyfill/issues/11. Happy to contribute to that one as well.

Not sure about what extra tests are needed:
- single signal canceling multiple requests?

Also, that polyfill takes care of initial point 2. ([patching fetch](https://github.com/mo/abortcontroller-polyfill/blob/master/src/patch-fetch.js))

joaovieira:
Pushed some changes up in https://github.com/jmeas/fetch/pull/2.

> Added a test for reusing the same signal to cancel multiple fetch requests (e.g. https://developers.google.com/web/updates/2017/09/abortable-fetch#one_signal_many_fetches)
>
> With that I noticed we needed a better polyfill to handle multiple event listeners, so I added an AbortSignal polyfill. It's on a separate file so it can be swapped with any other implementation. I'm keen on fixing mo/abortcontroller-polyfill#11 and use that instead when ready.

joaovieira:
As mentioned above, I've fixed and added [abortcontroller-polyfill](https://github.com/mo/abortcontroller-polyfill) as a dev dependency for tests (instead of own implementation):
* https://github.com/mo/abortcontroller-polyfill/pull/12
* https://github.com/mo/abortcontroller-polyfill/pull/13
* https://github.com/mo/abortcontroller-polyfill/pull/14 (this one will fix Travis)

joaovieira:
Looking back at what @jimmywarting mentioned in https://github.com/github/fetch/pull/592#issuecomment-350062956. Request does have a default `AbortSignal` ([Request spec](https://fetch.spec.whatwg.org/#request-class) and [web-platform-tests](https://github.com/w3c/web-platform-tests/pull/6484#issuecomment-315775251)). I.e.:

```js
// Both Firefox and Edge
const req = new Request('https://google.com');
console.log(req.signal); // AbortSignal { aborted: false, onabort: null }
```

The reasoning may be because the `request.signal` should actually proxy events from the original signal (without being the same):

> All requests have a signal property whether one is provided in the constructor or not. When a signal is given to the constructor, request.signal listens to its signals, and mimics them. This means request.signal never equals the signal passed into the constructor.

This would mean we'd need to bundle it here. What do you think? Seems a tiny use case imo. When we'll have signal subclasses and that becomes a problem, we probably won't be needing a polyfill anymore :)

jmeas:
:wave: Hi again! Thank you for doing so much good work here, @joaovieira !

I have some more time now to look back at this PR, if there's anything else I can do. @mislav, what do you think?

README.md
```diff
@@ -260,6 +261,36 @@ response.headers['X-Request-URL'] = request.url
...
+
+This polyfill supports
+[the abortable fetch API](https://developers.google.com/web/updates/2017/09/abortable-fetch).
+However, aborting a fetch requires use of two additional DOM APIs:
+[AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)
+and
+[AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal).
+Typically, browsers that do not support fetch will also not support
+AbortController or AbortSignal. Consequently, you will need to include an
+additional polyfill for these APIs to abort fetches.
```
mislav: If there's a polyfill that we audited and recommend using, then we should link to it here.

fetch.js
```diff
@@ -443,6 +468,10 @@
         reject(new TypeError('Network request failed'))
       }
 
+      xhr.onabort = function() {
```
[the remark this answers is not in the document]
joaovieira: @mislav I believe this is the most correct as mentioned [here](https://github.com/github/fetch/pull/572#issuecomment-337740989). Calling abort on the XHR will set the right status code and `readyState` (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort).

fetch.js
```diff
@@ -459,6 +488,17 @@
         xhr.setRequestHeader(name, value)
       })
 
+      if (request.signal) {
+        request.signal.addEventListener('abort', abortXhr)
+
+        xhr.onreadystatechange = function() {
```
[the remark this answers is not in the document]
joaovieira: this is the cleanest/safest way to know when the request has finished, regardless of the result, instead of having to do so on every callback:

```js
xhr.onload = function() {
  ...
  request.signal.removeEventListener('abort', abortXhr)
}

xhr.onabort = function() {
  ...
  request.signal.removeEventListener('abort', abortXhr)
}

xhr.onerror = function() {
  ...
  request.signal.removeEventListener('abort', abortXhr)
}

xhr.onprogress = function() {
  ...
  // should I remove listener? no.
}

// what other events there are? timeout?
...
```

having the `removeListener` close to `addListener` is also easier to read/maintain.

joaovieira:
@mislav agree that they should go together but done on separate PRs as you mentioned in the beginning. I guess we keep it open and merge both at the same time?

We can start by looking into [abortcontroller-polyfill/patch-fetch](https://github.com/mo/abortcontroller-polyfill/blob/master/src/patch-fetch.js) as it's already doing that. Moving forward, I'd prefer to see that polyfill only implementing and exporting `AbortController`/`AbortSignal` and not patching, but that needs to be discussed with @mo.

fetch.js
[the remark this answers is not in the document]
mislav: I'm fine with `xhr.abort()`. I'm just saying that we don't need to install the `xhr.onabort` _callback_ just to reject the promise. We know that `abortXhr()` will be executed, so we can reject the promise in there.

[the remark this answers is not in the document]
joaovieira: I see.. xhr is never exposed so it can't be aborted from the outside. Is there any other way it could be aborted by the browser? If not, there shouldn't be any difference. I don't mind this way though, reads well :)

jamesplease:
> This is a good candidate for another PR, but what would be your idea how to proceed with that? Let's make a rough plan before we merge this.

I agree with @joaovieira that linking out to the `abortcontroller-polyfill` library seems best. What do you think?

fetch.js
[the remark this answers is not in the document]
mislav: `onabort` wasn't needed until now, and since we control literally the only place from which the xhr is aborted, I don't think adding one level of indirection on top of `abortXhr` is necessary.

mislav:
> I agree with @joaovieira that linking out to the `abortcontroller-polyfill` library seems best.

I also agree that since we already need AbortController polyfill anyway, we can just tell our users to use it in a way that also patches native `window.fetch` and Request objects.

This should be a small enough documenation change that it should be appropriate for this PR. Does someone care to write up a section in the README about aborting? After that this will be ready for merge.

jamesplease:
@mislav I will do that now.

jamesplease:
Hmmm, actually, I just woke up and realized I wasn't putting enough thought into this.

There are three situations we need to account for:

1. browser has neither AbortController nor fetch
2. browser has no AbortController, but it does have fetch
3. browser has both

And there are two options for how we load the polyfills:

```js
import 'whatwg-fetch';
import 'abortcontroller-polyfill';
```

or

```js
import 'abortcontroller-polyfill';
import 'whatwg-fetch';
```

`abortcontroller-polyfill` does nothing if fetch doesn't exist, so it seems like we need the order of things to go:

```js
import 'whatwg-fetch';
import 'abortcontroller-polyfill';
```

but I would like to verify that this works as expected in the three situations listed above. I'm about to catch the shuttle but I should have time to do this today. If someone else does it before me and wants to jump in with updating the documentation, then go for it!

joaovieira:
Ok, let me clarify my points:

1. I meant to pull parts of that [patch-fetch.js](https://github.com/mo/abortcontroller-polyfill/blob/master/src/patch-fetch.js) into here as another PR. This is to tackle part 2. of the problem identified in https://github.com/github/fetch/pull/592#pullrequestreview-81886705. Since we just added the aborting behaviour for the polyfilled fetch, it sounds natural to patch the native fetch in here as well. As a user I'd expect that.

2. If I hadn't seen the source code of `abortcontroller-polyfill`, I would not expect it to patch fetch. I'd assume it just provides an implementation of `AbortController`/`AbortSignal`. As @jimmywarting mentioned in https://github.com/github/fetch/pull/572#issuecomment-341198011, `AbortController` is not fetch specific.

In this scenario, the import order doesn't matter since `abortcontroller-polyfill` would not do anything to `window.fetch` and `whatwg-fetch` does not require `AbortController`. This is also what @just-boris mentioned in https://github.com/github/fetch/pull/572#issuecomment-337740989.

fetch.js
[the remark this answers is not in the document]
dysonpro: Doesn't `onAbort()` also confirm that `xhr.abort()` was successful vs. just trusting the results of `abortXhr()`?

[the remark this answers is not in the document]
mislav: @dysonpro That's correct, but we aren't interested in whether `xhr.abort()` was performed successfully or not. We want to reject the promise right away because we know for sure an abort was requested.

mo:
I have released abortcontroller-polyfill v1.1.8 where you can do ```import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only'``` to polyfill AbortController/AbortSignal without patching fetch.

mislav:
@mo Thank you! What would you recommend for patching everything (i.e. bring in AbortController polyfill & fetch polyfill & abortable fetch polyfill) without having to load files that bring in AbortController polyfill twice?

So far I can only come up with something like:

```js
import './abortcontroller-polyfill'
import './fetch-polyfill'

// abortcontroller-polyfill.js
import {AbortController, AbortSignal} from 'abortcontroller-polyfill/dist/cjs-ponyfill'
if (!window.AbortController) {
  window.AbortController = AbortController
  window.AbortSignal = AbortSignal
}

// fetch-polyfill.js
import 'whatwg-fetch'
import {abortableFetch} from 'abortcontroller-polyfill/dist/cjs-ponyfill'
const {fetch, Request} = abortableFetch(window)
window.fetch = fetch
window.request = Request
```

But it feels a little bit manual.

mo:
I was thinking that the end user would do:

```js
import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only'
import 'whatwg-fetch'
```

The first line would install the AbortController/AbortSignal on the global object and the latter would install a version of fetch that supports aborting. In this case the "fetch patching" code in abortcontroller-polyfill would not be used at all, and whatwg-fetch would be able to do a proper xmlHttpRequest.abort(). I'm thinking this will not load AbortController twice?

mislav:
> I'm thinking this will not load AbortController twice?

What if the browser already implements native `fetch` (all modern browsers nowadays) but no aborting?

mo:
I was thinking that whatwg-fetch would detect that the native fetch lacks abort support and install itself anyway. This would require changes on your end of course, but it is the only way that the request could be properly aborted.

You can of course install the polyfills in the other order today, but then you don't get 100% spec compliant behavior because you get "the response of the request is ignored" rather than "the tcp connection is terminated when abort is called". By "installing the polyfills in the other order" I just mean this:
https://raw.githubusercontent.com/mo/abortcontroller-polyfill-examples/master/plain-javascript-github-fetch/index.html

mislav:
> I was thinking that whatwg-fetch would detect that the native fetch lacks abort support and install itself anyway. This would require changes on your end of course, but it is the only way that the request could be properly aborted.

That's true. However, I'm not sure that gaining abort functionality warrants overriding the native implementation. I'll have to discuss this with others on my team to see what the general consensus around such a change would be. Thank you for your input!

fetch.js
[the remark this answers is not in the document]
dysonpro: Thanks @mislav for your explanation. I am excited about this functionality so I have been following this PR carefully. I am already using this branch code in a project that requires aborting no longer needed requests.
Your point is that you want to ensure the promise is rejected, correct? I feel that it is closer to the core intention of the PR (and my own project as well) to ensure that the abort was completed successfully. Is there a way to do that without utilizing the `xhr.onAbort()` callback? Without that I feel that there could be a gap in the verification of the functionality for every platform and environment. Thanks again for your feedback.

test/test.html
```diff
@@ -33,6 +33,7 @@
   </script>
 
   <script src="/node_modules/promise-polyfill/promise.js"></script>
+  <script src="/node_modules/abortcontroller-polyfill/dist/abortcontroller.js"></script>
```
mislav: This can now import `abortcontroller-polyfill/dist/abortcontroller-polyfill-only` since v1.1.8

fetch.js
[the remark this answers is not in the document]
mislav: @dysonpro Your concerns are valid, but we have different priorities when thinking about this. At the point when `abortXhr()` is invoked (which can only happen upon AbortSignal), I consider the request most definitely discarded. We can then call `xhr.abort(); reject(...)` all at once. Now, if there is a browser which would somehow fail to actually abort the XHR when `xhr.abort()` is called, that is fine with me because we already ensured that the promise is rejected. I don't expect this case to happen often (or ever, for that matter) and so I'm not concerned with it. The code change that I suggested, after all, is mostly stylistic.

mislav:
I have discussed overriding native `fetch` to gain abort functionality with some on my team and here's a compromise we came down to: we will avoid ever overriding native `fetch`, but we will make module exports consumable in a way such as `import {fetch, Request} from 'whatwg-fetch'` that allows you access to our own implementations even when browser native equivalents were available. That way, if you use `fetch` as exported from our module and you have already polyfilled AbortController/Signal, you will always have access to truly abortable requests no matter the browser, because you will be using our own XHR implementation.

mo:
So to get "abort means tcp socket is closed" polyfilled globally you would do:

```js
import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only'
import {fetch, Request} from 'whatwg-fetch'
window.fetch = fetch;
window.Request = Request;
```

But if you did the following instead you'd get "abort means tcp socket is closed" on non-fetch browsers but "abort means promise is rejected and response is ignored but tcp socket is not closed" for browsers with non-abortable fetch:

```js
import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only'
import 'whatwg-fetch'
```

And people who upgrade "whatwg-fetch" without reading release note and thus keep just importing "whatwg-fetch" alone like they did before; these people will get the same behavior as in the case above so it will work for them too (it's only if you really want "proper abort" that you need to modify you imports to fit the first case):

```js
import 'whatwg-fetch'
```

@mislav did I understand that correctly?

mislav:
@mo I would not suggest ever overriding `window.fetch` in one's application, since there's no need when someone uses a module system. Here's what I'm suggesting:

```js
// Use browser native fetch or polyfilled fetch for older browsers. If the browser
// implements fetch, you are at the mercy of its implementation. Using abort
// functionality would not be safe.
import 'whatwg-fetch'
window.fetch('/').then(...)

// Use polyfilled fetch regardless of browser implementation existing or not. This
// buys you consistent behavior across browsers, which is useful for accessing
// newer features such as aborting.
import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only'
import {fetch, Request} from 'whatwg-fetch'
fetch(new Request('/', { signal: controller.signal })).then(...)
```

Yes, the latter approach is a bit more manual and more conscious, but we'd like to take the conservative approach with our polyfill and not ever override browser's native implemenation.

mo:
Sounds excellent.

dysonpro:
@mislav Is that going to be a part of the same PR? I'm looking forward to taking advantage of the great changes so far!

mislav:
@dysonpro First we are going to set up the exports (currently `whatwg-fetch` is not a module) in a separate PR, and then we are going to merge this PR.

joaovieira:
Hm, since you don't want to patch fetch in `whatwg-fetch`, then wouldn't this be enough without this PR?:
```js
// Option 1. Support for older browsers, conservative fetch polyfill (import order matters)
import 'whatwg-fetch'; // as-is
import 'abortcontroller-polyfill'; // patches native or polyfilled fetch with aborting

// Option 2. No need to polyfill fetch (you're only targeting modern browsers)
import 'abortcontroller-polyfill'; // patches native fetch with aborting
```

Though what I was looking for was:
```js
// Option 1. Support for older browsers, conservative fetch polyfill (import order does not matter)
import 'abortcontroller-polyfill'; // no patch
import 'whatwg-fetch'; // polyfills whole fetch or patch native fetch with aborting

// Option 2. No need to polyfill fetch (you're only targeting modern browsers)
import 'abortcontroller-polyfill'; // no patch
import 'abortcontroller-polyfill/fetch-signal-polyfill'; // patches native fetch with aborting
```

And ultimately:
```js
// Option 1. Support for older browsers + node (import order does not matter)
import 'abortcontroller-polyfill'; // no patch
import 'isomorphic-fetch'; // both whatwg-fetch and node-fetch polyfill whole fetch or patch native fetch with aborting
```

or

```js
// Option 1. Support for older browsers + node (import order matters)
import 'isomorphic-fetch'; // no patch
import 'abortcontroller-polyfill'; // no patch
import 'abortcontroller-polyfill/fetch-signal-polyfill'; // patches native or polyfilled fetch with aborting
```

All is safe and transparent to the user - i.e. no need to choose between module exports vs side-effects, know what is safe and not safe and/or know how to patch fetch with an imperative API.

mislav:
@joaovieira Your examples are valid. However, in this project, as well as our documentation and examples, we've resolved to avoid any venue where native `window.fetch` or `window.Request` are ovewritten. This comes at a cost of users not being able to use latest spec features in a transparent way, as you pointed out, but comes at a benefit of clearer intents when reading code and debugging.

jayphelps:
I have cycles to push this over the finish line if there's anything I can PR? 😄Happy to help!

mislav:
@jayphelps Well, basically what we need to do right now (in a separate PR) is make `fetch.js` be a UMD module and declare its exports. Regardless of the module stuff, it needs to keep polyfilling `window.fetch` to keep backwards compatibility. If you want to help set this up, PR welcome! ✨

jayphelps:
I’m on it 👍

phil-lgr:
Let us know if we can help in any way, I think aborting request is a pretty important feature to have!

fetch.js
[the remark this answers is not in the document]
MattiasBuelens: > Is there any other way it could be aborted (e.g. by the browser)? If not, there shouldn't be any difference.

* A browser extension (e.g. an ad blocker) could abort a request.
* A service worker could intercept the fetch and replace it with a new request that has (or will have) an aborted signal.

Arguably, these are rather obscure use cases, but they're real nonetheless. I think a general `fetch` polyfill should handle these, so it should listen for `xhr.onabort`.

fetch.js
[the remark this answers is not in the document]
MattiasBuelens: Another possibility is [`xhr.onloadend`](https://xhr.spec.whatwg.org/#event-xhr-loadend). This event is dispatched after the request completes, regardless of *how* completes (`load`, `error`, `abort` or `timeout`).

jayphelps:
Hey everyone 👋This Dot and Netflix have partnered up and released a fork of this project called [yetch](https://github.com/Netflix/yetch) which adds support for CJS exports, as well as supporting AbortController signals. The main package import `import { fetch } from 'yetch'` or `import 'yetch'` does not polyfill the globals (so that you can use them as CJS without taking over the globals). If you just want to polyfill, use `import 'yetch/polyfill'`.

e.g.

```js
import 'abortcontroller-polyfill';
import 'yetch/polyfill';
```

And then can be used just like the real deal:

```js
const controller = new AbortController();
const url = 'https://reqres.in/api/users?delay=5';

fetch(url, { signal: controller.signal }).then(response => {
  return response.text();
}).then(text => {
  console.log(text);
}).catch(err => {
  if (err.name === 'AbortError') {
    console.log('Fetch aborted');
  } else {
    console.error('Uh oh, an error!', err);
  }
});

setTimeout(() => {
  controller.abort();
}, 500);
```

***

We totally realize that OSS is often a volunteer labor of love and sometimes maintainers must deprioritize--we're maintainers too, we've been there. So we don't wish to imply any disrespect for the maintains of whatwg-fetch, or GitHub and hope our two projects can rejoin when the time permits. ✌️

mislav:
Please don't advertise your projects on our issue tracker. 🙇

mislav:
Thank you for your patience!

jackywq:
@joaovieira， thank you for your advice, I have solved the problem!

jamesplease:
Thank you @mislav @jayphelps @joaovieira @jimmywarting and everyone else!

newmanw:
Any idea on when the next release will be pushed?  I see this was merged but doesn't look like there has been a release in a while.

Issue #594, opened 2017-12-16
Testing fails with chai v4.x


ghisvail:
See [Debian-Bug#884154](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=884154)

chrismusselman:
i think the  comments are key to success

ghisvail:
> i think the comments are key to success

What do you mean?

arunasank:
I don't see any tests failing, when I upgrade chai to 4.1.2 on https://github.com/github/fetch/pull/597. What tests are failing?

ghisvail:
@arunasank Please read the link to the Debian bug report. The complete log is provided.

arunasank:
```
arunasank:fetch aruna$ npm test

> whatwg-fetch@2.0.3 test /Users/aruna/git/fetch
> make

./node_modules/.bin/jshint *.js test/*.js
./script/test


  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․ArrayBuffer is deprecated in XMLHttpRequest.send(). Use ArrayBufferView instead.
․․․․․․․․․․․․․․․․․․․․․․․․․

  118 passing (1s)
  9 pending

ArrayBuffer is deprecated in XMLHttpRequest.send(). Use ArrayBufferView instead.


  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․

  112 passing (21ms)
  15 pending
```

@ghisvail ok, thank you for that. The above is the output when I run the tests on my mac. Why are some tests still in `pending` state? How can I run these?

mislav:
In #626 we've upgraded Mocha and Chai and all tests pass in all supported browsers.

Please provide more info if something fails. Note that the testing setup has been reconfigured from the ground up. See https://github.com/github/fetch/blob/master/CONTRIBUTING.md

PR #597, opened 2018-01-19
Chai 4.1.2

arunasank:


arunasank:
Fixes https://github.com/github/fetch/issues/594

package.json
```diff
@@ -19,5 +18,8 @@
   ],
   "scripts": {
     "test": "make"
+  },
+  "dependencies": {
+    "chai": "^4.1.2"
```
dgraham: Updating the version looks good. This should still be listed as a `devDependency`, though.

mislav:
This was done in #626.

PR #609, opened 2018-03-02
0.11

dapen666:


jaafarexcz:
> In Safari V9.0.3 (10601.4.4), neither `then()` nor `catch()` is called when a request timeout occurs. Instead, Safari prints out the following error to the console:35896c5106a8f8aef8b6357de8df2c98b4168c4d
> 
> screen shot 2016-03-11 at 3 43 22 pm
> 
> In Chrome V49.0.2623.87 (64-bit), the `catch()` callback is properly executes as expected.
> 
> Test Script Used:
> 
> ```
> fetch('http://10.255.255.1') // a non-routable IP address
>     .then(function(response) {
>         console.log("response.ok", response.ok);
>     }).catch(function(error) {
>         console.log('Failed: ', error)
>     });
> ```
> Without either `then()` or `catch()` being called on timeout, how are we supposed handle this type of request failure for Safari browsers?

9

jaafarexcz:
> _No description provided._
35896c5106a8f8aef8b6357de8df2c98b4168c4d

PR #616, opened 2018-03-29
build/distribute as UMD

jayphelps:
~~I still need to adjust tests--still trying to figure out if I can change things while still keeping the existing makefile setup, is that the preference?~~ Ready!

This was requested before #592 can be merged.


package.json
```diff
@@ -11,13 +11,17 @@
...
     "promise-polyfill": "6.0.2",
-    "url-search-params": "0.6.1"
+    "url-search-params": "0.6.1",
+    "webpack": "^4.4.1",
+    "webpack-cli": "^2.0.13"
   },
   "files": [
     "LICENSE",
-    "fetch.js"
+    "dist/whatwg-fetch.js"
```
jayphelps: If it's desired we could build to `./fetch.js` instead of putting things inside `dist` (and naming it something else). I did it this way as it's more traditional in the community, but otherwise your call.

webpack.config.js
```diff
@@ -0,0 +1,12 @@
+const path = require('path');
+
+module.exports = {
+  mode: 'development',
+  entry: './src/index.js',
+  output: {
+    filename: 'whatwg-fetch.js',
+    path: path.resolve(__dirname, 'dist'),
+    library: 'WHATWGFetch',
```
jayphelps: Wasn't sure what to call the global object. Any preferences?

src/index.js
```diff
@@ -0,0 +1,476 @@
+'use strict';
```
jayphelps: The way github is displaying this isn't great (it doesn't realize the file move/rename and changes). Here is the separate commit with whitespace ignored (`?w=1` flag) https://github.com/github/fetch/pull/616/commits/a5ec27348af874916d860f611b6f166aaad9ae5a?w=1#diff-1fdf421c05c1140f6d71444ea2b27638

mislav:
Thank you for working on this! What do you think about skipping webpack dependency altogether (it might be an overkill), and wrap current `fetch.js` with something like this:

```js
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
        define([], function(){ return factory({}, root) })
    } else if (typeof module === 'object' && module.exports) {
        factory(module.exports, root)
    } else {
        factory(root, root)
  }
}(typeof self !== 'undefined' ? self : this, function (exports, global) {

    // current fetch.js contents

    exports.fetch = fetch
    exports.Request = Request
    // ...
    return exports
}));
```

fetch.js
```diff
@@ -1,23 +1,29 @@
-(function(self) {
+(function (root, factory) {
+  /* jshint strict:false */
+  /* globals define, module */
```
jayphelps: I initially tried to use`/* jshint ignore: start */` (and end) but it didn't work in 2.8.0 jshint (maybe https://github.com/jshint/jshint/issues/2411). After upgrading to jshint 2.9.0 it works, but then it says factory is unused, and everything I tried to ignore that caused even more issues, and 2.9.0 also detects other undefined variables like Promise/Symbol. So this seemed to be the solution with the least amount of changes.

jayphelps:
@mislav based on our PM conversations I've updated the PR to have the UMD preamble inline. I also added a simple test to verify the exports are there in CJS. ✌️

fetch.js
```diff
@@ -1,23 +1,29 @@
-(function(self) {
+(function(root, factory) {
+  /* jshint strict:false */
+  /* globals define, module */
+  if (typeof define === 'function' && define.amd) {
```
jayphelps: I initially tried to use/* jshint ignore: start */ (and end) but it didn't work in 2.8.0 jshint (maybe jshint/jshint#2411). After upgrading to jshint 2.9.0 it works, but then it says factory is unused, and everything I tried to ignore that caused even more issues, and 2.9.0 also detects other undefined variables like Promise/Symbol. So this seemed to be the solution with the least amount of changes.

fetch.js
```diff
@@ -1,23 +1,29 @@
...
 
-  if (self.fetch) {
-    return
-  }
-
   var support = {
-    searchParams: 'URLSearchParams' in self,
-    iterable: 'Symbol' in self && 'iterator' in Symbol,
-    blob: 'FileReader' in self && 'Blob' in self && (function() {
+    searchParams: 'URLSearchParams' in global,
```
mislav: To keep the diff to a minimum, could `global` simply continue to be called `self` like before?

script/test
```diff
@@ -7,3 +7,5 @@ if [ -n "$SAUCE_BROWSER" ]; then
 else
   ./script/phantomjs
 fi
+
+./node_modules/mocha/bin/mocha test/cjs.js
```
mislav: Why is this separate file needed?

These tests could be in `test.js`. That file [already jumps through hoops](https://github.com/github/fetch/blob/cbb313bf09512efda9444e40fe9e8ac432ac1108/test/test.js#L83-L95) to exercise both native and polyfilled fetch, and this module approach could potentially clean it up.

jayphelps: A separate file was used as it is testing whether the CJS exports are available and the easiest and foolproof way seemed to be to do it inside a real CJS environment. test.js is evaluated in the context of a browser, where no `require` exists, and I did not see any other tests running inside a CJS env. Sorry about that, am I missing an easier solution?

fetch.js
jayphelps: done

script/test
jayphelps: @mislav any thoughts? 🕺

phil-lgr:
Hey let us know if we can help here! This is a much needed feature!

graingert:
fyi https://www.npmjs.com/package/fetch-ponyfill already provides a module for this

jayphelps:
Hey everyone 👋This Dot and Netflix have partnered up and released a fork of this project called [yetch](https://github.com/Netflix/yetch) which adds support for CJS exports, as well as supporting AbortController signals. The main package import `import { fetch } from 'yetch'` or `import 'yetch'` does not polyfill the globals (so that you can use them as CJS without taking over the globals). If you just want to polyfill, use `import 'yetch/polyfill'`.

e.g.

```js
import 'abortcontroller-polyfill';
import 'yetch/polyfill';
```

And then can be used just like the real deal:

```js
const controller = new AbortController();
const url = 'https://reqres.in/api/users?delay=5';

fetch(url, { signal: controller.signal }).then(response => {
  return response.text();
}).then(text => {
  console.log(text);
}).catch(err => {
  if (err.name === 'AbortError') {
    console.log('Fetch aborted');
  } else {
    console.error('Uh oh, an error!', err);
  }
});

setTimeout(() => {
  controller.abort();
}, 500);
```

***

We totally realize that OSS is often a volunteer labor of love and sometimes maintainers must deprioritize--we're maintainers too, we've been there. So we don't wish to imply any disrespect for the maintains of whatwg-fetch, or GitHub and hope our two projects can rejoin when the time permits. ✌️

script/test
jayphelps: bump

stefanpenner:
@mislav anything I can do to help get this, or something similar landed? I would hate to have rely on a fork long-term.

mislav:
Hi @jayphelps, I've just pushed my proposed changes to this branch. Thank you for waiting.

Everything was OK with the state of the PR as you've left it. The static UMD wrapper looked good. However, on our side, some new requirements to how we package our open source libraries have crept in. Specifically, we want our packages to support a `module` keyword in addition to the `main` keyword in `package.json`. The `module` file would be an entrypoint that is an ES5 file with ES6 export syntax. This is compatible with build tools such as Rollup.

I have made a call to have `fetch.js` be a plain, unwrapped implementation of the polyfill with ES6 exports, therefore suitable to be pointed to by `module`. This would mean, however, that it would be best to have a build step that converts that into a UMD file, like you've first proposed in the initial version of this PR. Instead of Webpack, I use Rollup here because its what we use in our other open source projects (sometimes we use Babel to generate the UMD file if the project already uses Babel).

Because jshint was giving me trouble through this transition, I've switched to eslint & prettier like we also do in our other open source projects. This resulted in a lot of apparent code changes that are mostly just prettier adjusting whitespace. This make the diff of this PR look huge, but I think the transition is worth it in the long run.

I would appreciate a review. Thank you! 🙏

stefanpenner:
>  I've switched to eslint & prettier like we also 

Good call!

jayphelps:
LGTM :shipit:

jayphelps:
A somewhat related offshoot concern I still have is that the current proposed abort support in #592 won't polyfill `window.fetch` if `window.fetch` is already assigned but doesn't itself support aborting. My gut says that would be non-obvious to people and potentially cause an issue where aborting works as expected in one browser but doesn't in another. With this PR adding modules, is there potential to reconsider #592's behavior or alternatively provide a way to opt-in to it always polyfilling if something isn't 100% supported? e.g. it would override `window.fetch` if any only if it doesn't support abort signals. If that as the default behavior isn't acceptable, perhaps an opt-in module? `import 'whatwg-fetch/polyfill-with-abort';`

The approach we took in yetch was that the default import/export of the project didn't polyfill anything so it could be used as a library without side effects. If they wanted to instead polyfill you use `import import 'yetch/polyfill';`. My intent isn't to advertise yetch, just using it as a reference.

graingert:
Ponyfills like yetch are always the preferred option

keithamus:
@jayphelps I do not disagree - but I think this is a good first step. Let's wait for this patch to land and then look into those subsequent issues.

mislav:
> The approach we took in yetch was that the default import/export of the project didn't polyfill anything so it could be used as a library without side effects.

There's definitely value in the ponyfill approach, especially in the world where using modules is becoming more and more standard. However, you have to consider the existing `whatwg-fetch` user base which is very wide and far-reaching in terms of history. We've advertised this as a polyfill from the start and I would guess with relative confidence that a large slice of our user base is simply including `fetch.js` in their project and using `window.fetch` after that.

I'm not ready to pull the plug on this behavior and make the polyfill an opt-in just yet, even through a major version bump, _especially_ because this library didn't provide exports in the past. So, my plan is to first provide the exports (while keeping the implicit polyfill functionality), have our users get used to that, and possibly switch to the ponyfill approach months down the line through another major version bump.

Thanks for your input!

wasem1772:
@github-cla

PR #620, opened 2018-04-20
mention isomorphic-fetch package

pravdomil:
see diff


graingert:
probably better to mention fetch-ponyfill, it doesn't mutate globals

dgraham:
See previous discussion in #316.

PR #626, opened 2018-05-24
Switch test suite from PhantomJS to Karma

mislav:
`npm t` now runs the test suite in headless Chrome/Firefox.

With `npm run karma`, additional browsers may be connected to the Karma runner and automatically re-run tests when files change.


.travis.yml
```diff
@@ -2,9 +2,15 @@ sudo: false
 language: node_js
 node_js:
   - "node"
+dist: trusty
+addons:
+  chrome: stable
+before_script:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
```
koddsson: Do we still need `xfvb` if we are using a headless version of Firefox/Chrome?

keithamus: I believe so

koddsson: I looked it up and [developers.google.com](https://developers.google.com/web/updates/2017/04/headless-chrome) says:

>So I still need Xvfb?

> No. Headless Chrome doesn't use a window so a display server like Xvfb is no longer needed. You can happily run your automated tests without it.

> What is Xvfb? Xvfb is an in-memory display server for Unix-like systems that enables you to run graphical applications (like Chrome) without an attached physical display. Many people use Xvfb to run earlier versions of Chrome to do "headless" testing.

keithamus: We still have it in https://github.com/chaijs/type-detect/blob/master/.travis.yml and I cannot remember why, but I'm sure I would have removed it if it was no longer needed. Perhaps Firefox still needs it? I'd be happy to be wrong though 😛

koddsson: I think the headless browsers will work without `xvfb`. It should be easy enough to test 😄

koddsson: [I made a PR](https://github.com/chaijs/type-detect/pull/136) to remove that pre-test step in `type-detect` and it's running 🍏

.travis.yml
```diff
@@ -2,9 +2,12 @@ sudo: false
 language: node_js
 node_js:
   - "node"
+dist: trusty
```
dgraham: Trusty is the default test environment now, so we don't need to specify it here.

I've needed `sudo: true` for Chrome headless to work in the past. That issue may be fixed now, though.

PR #652, opened 2018-09-03
Update Karma and detect available browsers when testing

mislav:


test/karma.config.js
```diff
@@ -3,7 +3,13 @@ const serverEndpoints = require('./server')
 module.exports = function(config) {
   config.set({
     basePath: '..',
-    frameworks: ['mocha', 'chai'],
+    frameworks: ['detectBrowsers', 'mocha', 'chai'],
+    detectBrowsers: {
+      preferHeadless: true,
+      usePhantomJS: false,
+      postDetection: availableBrowsers =>
+        availableBrowsers.map(browser => (browser.startsWith('Chrom') ? `${browser}NoSandbox` : browser))
```
keithamus: is `Chrom` intentional here?

mislav:
This change causes a bizarre CI failure:

```
HeadlessChrome 0.0.0 (Ubuntu 0.0.0) native Request credentials defaults to same-origin FAILED
	AssertionError: expected 'omit' to equal 'same-origin'
```

Which would suggest that the version of Chrome used was 62, not 68. This branch uses the same karma-chrome-launcher version as master, so I can't guess right now why would the result be different.

PR #672, opened 2018-10-31
Update package.json

priyanksharma477:


keithamus:
Thanks for this PR @priyanksharma477 but it's a small inconsiquential change that I don't think we need. I'll close this for now.

PR #715, opened 2019-09-17
Provide Window Global to Verify Polyfill Load

mfasman95:
At the company I work for we leverage integration tests to verify our browser compatibility. One portion of that is verifying that we have properly loaded our polyfills. It would be beneficial to have some kind of window global variable that allowed us to determine if the polyfilling script was loaded on the page (whether or not the polyfill was necessary).

This feature is currently in `@babel/polyfill`:
```js
window._babelPolyfill; // true
```
It is also available in `core-js`:
```js
window.core.version; // "{major}.{minor}.{patch}" - a semver string
```

I do not care what the variable is called or how the result is formatted - I would just like some way to verify that this polyfilling script has loaded.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -514,3 +514,5 @@ if (!self.fetch) {
   self.Request = Request
   self.Response = Response
 }
+
+self.WHATWGFetch = true
--- a/fetch.js
+++ b/fetch.js
@@ -514,5 +514,3 @@ if (!self.fetch) {
   self.Request = Request
   self.Response = Response
 }
-
-// self.WHATWGFetch = true
--- a/fetch.js
+++ b/fetch.js
@@ -515,4 +515,4 @@ if (!self.fetch) {
   self.Response = Response
 }
 
-self.WHATWGFetch = true
+// self.WHATWGFetch = true
--- a/fetch.js
+++ b/fetch.js
@@ -515,4 +515,4 @@ if (!self.fetch) {
   self.Response = Response
 }
 
-self.WHATWGFetch = true
+self._whatwgFetchPolyfill = true


mfasman95:
From what I can tell, those two test failures exist with or without my proposed change. I am not sure if I should do something to try and resolve the issues...?

JakeChampion:
You can do this already by calling `fetch.toString()` and confirming it is not `"function fetch() { [native code] }"`.

dasa:
Maybe check for this? https://github.com/github/fetch/blob/master/fetch.js#L509

mfasman95:
@JakeChampion and @dasa This will only be true if `fetch` was not present on the window. I am attempting to determine if the polyfill was loaded regardless of whether or not it was applied.

dasa:
I think it's always there on the exported `fetch` function.

Try:

```js
import {fetch as fetchPolyfill} from 'whatwg-fetch';
console.log(fetchPolyfill.polyfill);
```

mfasman95:
I tried doing this in the browser console in a window where the polyfill was loaded - the polyfill property is undefined.

If you look at the code carefully around the line you linked the property is being set on the fetch function, but that function is not being set to `self.fetch` unless `self.fetch` is falsy. This means that it will not be set on the window global fetch unless the polyfill version of the function is the one that is being used.

Basically, if what @JakeChampion suggested resolves to true then your suggestion will resolve to undefined.

mfasman95:
If that line was after the `if` check then maybe it would work?

dasa:
> This means that it will not be set on the window global fetch unless the polyfill version of the function is the one that is being used.

Agreed, but then you're not referring to the exported function. If you use the import, then you're sure to be accessing the shim, rather then checking if it decided to set itself on the window.

PR #729, opened 2019-10-20
Add typings

resynth1943:
I added some TypeScript type definitions so I can actually use this module :P

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1 @@
+export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -62,14 +62,6 @@ declare module 'whatwg-fetch' {
     redirect?: RequestRedirect
     signal?: AbortSignal | null
 
-    // node-fetch extensions
-    agent?: Agent | ((parsedUrl: URL) => Agent) // =null http.Agent instance, allows custom proxy, certificate etc.
-    compress?: boolean // =true support gzip/deflate content encoding. false to disable
-    follow?: number // =20 maximum redirect count. 0 to not follow redirect
-    size?: number // =0 maximum response body size in bytes. 0 to disable
-    timeout?: number // =0 req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
-
-    // node-fetch does not support mode, cache or credentials options
   }
 
   export type RequestContext =
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,4 @@
 declare module 'whatwg-fetch' {
-  import {Agent} from 'http'
-  import {URLSearchParams, URL} from 'url'
-
   export interface AbortSignal {
     aborted: boolean
 
@@ -41,9 +38,6 @@ declare module 'whatwg-fetch' {
     redirect: RequestRedirect
     referrer: string
     url: string
-
-    // node-fetch extensions to the whatwg/fetch spec
-    agent?: Agent | ((parsedUrl: URL) => Agent)
     compress: boolean
     counter: number
     follow: number
@@ -121,7 +115,7 @@ declare module 'whatwg-fetch' {
     values(): Iterator<[string]>
     [Symbol.iterator](): Iterator<[string, string]>
   }
-
+  
   type BlobPart = ArrayBuffer | ArrayBufferView | Blob | string
 
   interface BlobOptions {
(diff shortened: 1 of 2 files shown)


resynth1943:
That failing check has nothing to do with my PR, by the way

koddsson:
Hey @resynth1943 and thanks for the PR.

I believe that typescript includes fetch typings out of the box regardless if you are using the polyfill or not. Adding `'dom'` to the `'lib'` key in your `tsconfig.json` should solve this for you without us needing to define types in the polyfill codebase.

Reference: https://www.typescriptlang.org/docs/handbook/compiler-options.html

resynth1943:
Yeah, but you have to declare the module as exporting those symbols. I'll edit this PR to use the "fetch" from the TypeScript libraries, but you'll still need an export. Give me a few hours, and I'll take a look.

Regards,
Will

resynth1943:
How do you like *this*? I see no errors from Vim lol, so I guess that means everything's OK now?

MicahZoltu:
@koddsson You should only include `dom` `lib` if you actually are in an environment that has the full `dom` API.  If you are in NodeJS and you bring in this library, you do _not_ have the full `dom` API, so you should not add `dom` to your `lib` in `tsconfig.json`.

index.d.ts
```diff
@@ -0,0 +1,209 @@
...
+    method?: string
+    redirect?: RequestRedirect
+    signal?: AbortSignal | null
+
+    // node-fetch extensions
+    agent?: Agent | ((parsedUrl: URL) => Agent) // =null http.Agent instance, allows custom proxy, certificate etc.
+    compress?: boolean // =true support gzip/deflate content encoding. false to disable
+    follow?: number // =20 maximum redirect count. 0 to not follow redirect
+    size?: number // =0 maximum response body size in bytes. 0 to disable
+    timeout?: number // =0 req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
```
MicahZoltu: It looks like this was copied from `node-fetch` definitions.  I don't believe this library has any of the `node-fetch` extensions so they should not be included.

MicahZoltu: Same for all of the node-fetch extensions throughout.

resynth1943: Changing it now :D

index.d.ts
```diff
@@ -0,0 +1,209 @@
+declare module 'whatwg-fetch' {
+  import {Agent} from 'http'
+  import {URLSearchParams, URL} from 'url'
```
MicahZoltu: Once the node-fetch extensions get dropped, I bet you can delete these imports.  You should only need native JS types to create this declaration file.

index.d.ts
```diff
@@ -0,0 +1,209 @@
...
+    slice(start?: number, end?: number): Blob
+  }
+
+  export class Body {
+    constructor(body?: any, opts?: {size?: number; timeout?: number})
+    arrayBuffer(): Promise<ArrayBuffer>
+    blob(): Promise<Buffer>
+    body: NodeJS.ReadableStream
+    bodyUsed: boolean
+    buffer(): Promise<Buffer>
```
MicahZoltu: You should verify, but I don't think this library uses `Buffer` or `NodeJS.ReadableStream`.  I suspect this is an artifact of copying from `node-fetch` definition.

index.d.ts
resynth1943: Done.

JakeChampion:
> @koddsson You should only include `dom` `lib` if you actually are in an environment that has the full `dom` API. If you are in NodeJS and you bring in this library, you do _not_ have the full `dom` API, so you should not add `dom` to your `lib` in `tsconfig.json`.

The [third bullet-point in the README.md](https://github.com/github/fetch#read-this-first) says: 
This project doesn't work under Node.js environments. It's meant for web browsers only.

This project is for browser environments, which means you would need to configure TypeScript to bring in the DOM API, which includes the `fetch` typings.

koddsson:
> How do you like _this_?

Like I said I believe that TypeScript supports fetch typings natively so there should be any need for typings so the issue isn't the types at all.

With that being said, we don't plan on adding any type support to the fetch polyfill. They should exists in the type system already or as a 3rd party type module.

resynth1943:
@koddsson 
```
Cannot export 'fetch'. Only local declarations can be exported from a module.ts(2661)
```

Apparently you're *wrong*.

Since you're refusing to merge this, I'll try and get them merged into DefinitelyTyped, just so this wasn't a complete waste of time. 

These typings are a necessity for TypeScript users, and I'm more than happy to spend time with this.

Many thanks,
resynth1943

PR #742, opened 2019-12-06
Allow reading FormData body back from request.

### LINKED ISSUE #460: Parsing formData failed with 'could not read FormData body as text'
Example code:

```js
const request = new Request('whatever', {
    method: 'POST',
    body: new FormData(),
});

request.formData(); // failed
```

I think this problem is caused by `this.text()` statement in [fetch.js#L283](https://github.com/github/fetch/blob/master/fetch.js#L283).

cameron-martin:
The FormData object is cloned, since [the spec](https://fetch.spec.whatwg.org/#dom-body-formdata) says that it gets encoded to a stream and decoded again. This is consistent with the behaviour in browsers.

The stream is also consumed by doing this, so subsequent reads fail with a TypeError.

Fixes #460

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -287,6 +287,14 @@ function Body() {
 
   if (support.formData) {
     this.formData = function() {
+      if (this._bodyFormData) {
+        var rejected = consumed(this)
+        if (rejected) {
+          return rejected
+        }
+        return Promise.resolve(cloneFormData(this._bodyFormData))
+      }
+
       return this.text().then(decode)
     }
   }
@@ -363,6 +371,16 @@ function decode(body) {
       }
     })
   return form
+}
+
+function cloneFormData(formData) {
+  const cloned = new FormData()
+
+  formData.forEach(function(value, key) {
+    cloned.append(key, value)
+  })
+
+  return cloned
 }
 
 function parseHeaders(rawHeaders) {
--- a/test/test.js
+++ b/test/test.js
@@ -86,6 +86,16 @@ function readArrayBufferAsText(buf) {
   return chars.join('')
 }
 
+function encodeFormData(formData) {
+  var encoded = ''
+
+  formData.forEach(function(value, key) {
+    encoded += encodeURIComponent(key) + '=' + encodeURIComponent(value) + '&'
+  })
+
+  return encoded.substring(0, encoded.length - 1)
+}
+
 var preservedGlobals = {}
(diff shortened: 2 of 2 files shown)


cameron-martin:
These test failures seem to be unrelated to my PR, since they fail when I run them locally on master also.

fetch.js
```diff
@@ -365,6 +373,16 @@ function decode(body) {
   return form
 }
 
+function cloneFormData(formData) {
+  const cloned = new FormData()
+
+  formData.forEach(function(value, key) {
```
mislav: This relies on browser native FormData implementing `forEach`.

Keep in mind that this polyfill only takes effect in browsers that don't have `fetch` in the first place. These are:
- IE
- Edge < 14
- Firefox < 39
- Chrome < 40
- Safari 10.1

All of these older browsers do implement FormData, but they implement an older spec of it which [doesn't have introspection features](https://caniuse.com/#feat=mdn-api_formdata_entries) such as `get()` or `forEach`.

Therefore I don't think the feature you've added here would work at all. What do you think?

cameron-martin: Good point. Do you think that returning the uncloned FormData object would be a good compromise, even though it breaks the spec a little?

mislav: I don't see it as a good compromise just yet. Most of why I don't is that I don't understand the use case behind your PR. Why would we only partially implement the spec, returning the FormData instance in a browser that doesn't let you inspect that object? What does the person gain from having this API available?

cameron-martin: So my use case is using this in jsdom. Jsdom does support the full formdata API but does not have the Request and Response classes so I am using this library to provide these.

I have an API client that has the concept of middleware (with type signature `(next: (request: Request) => Promise<Response>) => (request: Request) => Promise<Response>`) and I'm wanting to test that a middleware has added the correct fields to the formdata object.

This is a slightly strange use case, admittedly, and there might be other ways of going about it, such as defining request and response types that are independent of any platform-specific apis, or testing this in real browsers.

mislav: Thank you for explaining your use-case. That all makes sense, but you are right that this is a strange case, and (in my opinion) it's not one we can reliably keep in mind while maintaining this library. (Lately we haven't been maintaining this library at all, since we no longer have a need for the polyfill on GitHub.)

cameron-martin: That's fair enough. I'll close this PR and use a different approach. Thanks for taking the time to respond.

PR #751, opened 2020-01-27
feat: add polyfill.js so as not to pollute global

steelydylan:
When using fetch, we don't want to pollute our client's global scope
This PR is for that
by doing like below we can safely use fetch polyfill without polluting global scope

```js
import { fetch } from 'whatwg-fetch/polyfill';
```

Thank you in advance.


dgraham:
Installing into the `window` global scope is intentional. This is designed as a polyfill that patches missing browser APIs rather than library. See the readme for a [description](https://github.com/github/fetch#importing) of the expected import behavior.

Issue #657, opened 2018-09-11
"self is not defined" error in React after upgrading to 3.0.0


karanmartian:
Hi, I just upgraded, but seeing this issue. Any suggestions?

``` 
/home/karan/Desktop/hbwapp/node_modules/whatwg-fetch/dist/fetch.umd.js:8
[0]     searchParams: 'URLSearchParams' in self,
[0]                                        ^
[0]
[0] ReferenceError: self is not defined
[0]     at /home/karan/Desktop/hbwapp/node_modules/whatwg-fetch/dist/fetch.umd.js:8:40
[0]     at support.searchParams (/home/karan/Desktop/hbwapp/node_modules/whatwg-fetch/dist/fetch.umd.js:2:66)
[0]     at Object.<anonymous> (/home/karan/Desktop/hbwapp/node_modules/whatwg-fetch/dist/fetch.umd.js:5:2)
[0]     at Module._compile (internal/modules/cjs/loader.js:702:30)
[0]     at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
[0]     at Module.load (internal/modules/cjs/loader.js:612:32)
[0]     at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
[0]     at Function.Module._load (internal/modules/cjs/loader.js:543:3)
[0]     at Module.require (internal/modules/cjs/loader.js:650:17)
[0]     at require (internal/modules/cjs/helpers.js:20:18)
```

mislav:
Hi could you tell us more about your module loader setup? Where does this happen, during compilation or execution in the browser? The more information the better, thank you!

davekedar:
I am using Expo for react native, after y'day all my new project breaks while loading. please help, this is due to upgrade from 2.0.4 to 3.0.0

davekedar:
What is the solution?

karanmartian:
@mislav mine is a React project and not React Native. This error happens at build time. I assume i need this polyfill for plain React projects

davekedar:
@mislav you are not getting the point, when we create the new project in Expo for react native it is fetching all the dependancies, thus this update is braking the initial loading. 

@karanmartian you can try restrict the version to 2.0.4 untill we get solution for this, it worked for me in react native

msafari:
I'm also seeing this exact error in our project which is a React project.

msafari:
ok, changing to the new `import { fetch as fetchPolyfill } from 'whatwg-fetch'`
fixed it.
[...]

PR #794, opened 2020-07-08
Fix https://github.com/github/fetch/issues/657

JakeChampion:
Fixes https://github.com/github/fetch/issues/657

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,551 +1,553 @@
-var support = {
-  searchParams: 'URLSearchParams' in self,
-  iterable: 'Symbol' in self && 'iterator' in Symbol,
-  blob:
-    'FileReader' in self &&
-    'Blob' in self &&
-    (function() {
-      try {
-        new Blob()
-        return true
-      } catch (e) {
-        return false
+(function(self) { 
+  var support = {
+    searchParams: 'URLSearchParams' in self,
+    iterable: 'Symbol' in self && 'iterator' in Symbol,
+    blob:
+      'FileReader' in self &&
+      'Blob' in self &&
+      (function() {
+        try {
+          new Blob()
+          return true
+        } catch (e) {
+          return false
+        }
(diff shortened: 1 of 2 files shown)


fetch.js
```diff
@@ -1,9 +1,13 @@
+var global = (function(self) {
+  return self
+  // eslint-disable-next-line no-invalid-this
+})(typeof self !== 'undefined' ? self : this)
```
jungdj, line 4: Can you please add some fallback such as an empty object?
I know this whatwg-fetch is intended to be used only in browser, however, it gives many conveniences by just "importing it" and not using it.

jungdj: To illustrate more, I'm suggesting to make importing 'whatwg-fetch' in nodejs environment works, while blocking the usage.

JakeChampion: This works in NodeJS correctly: Here the output of running it in NodeJS
```
❯ node -v
v14.4.0
❯ node
Welcome to Node.js v14.4.0.
Type ".help" for more information.
> var global = (function(self) {
...   return self
...   // eslint-disable-next-line no-invalid-this
... })(typeof self !== 'undefined' ? self : this)
undefined
> global
<ref *1> Object [global] {
  global: [Circular *1],
  clearInterval: [Function: clearInterval],
  clearTimeout: [Function: clearTimeout],
  setInterval: [Function: setInterval],
  setTimeout: [Function: setTimeout] {
    [Symbol(nodejs.util.promisify.custom)]: [Function (anonymous)]
  },
  queueMicrotask: [Function: queueMicrotask],
  clearImmediate: [Function: clearImmediate],
  setImmediate: [Function: setImmediate] {
    [Symbol(nodejs.util.promisify.custom)]: [Function (anonymous)]
  }
}
> 
(To exit, press ^C again or ^D or type .exit)
> 
```

Issue #801, opened 2020-07-22
Response.arrayBuffer() may resolve to ArrayBufferView


freund17:
```ts
new Response(new Uint8Array([8])).arrayBuffer();
```

returns a Promise, which resolves to a `Uint8Array`.
It should resolve to an `ArrayBuffer` instead.

Steps to reproduce:
Load this polyfill and paste above code into the developer console.

My hacky workaround:

After loading the polyfill add the following:

```ts
const ori = window.Response.prototype.arrayBuffer;

window.Response.prototype.arrayBuffer = function (...props) {
  return (ori.apply(this, props) as Promise<
    ArrayBuffer | ArrayBufferView
  >).then((value) => {
    if (ArrayBuffer.isView(value)) {
      return value.buffer.slice(
        value.byteOffset,
        value.byteOffset + value.byteLength
      );
    } else {
      return value;
    }
  });
};
```

PR #816, opened 2020-07-31
Make Response.arrayBuffer() always resolve with a `ArrayBuffer`

JakeChampion:
Fixes https://github.com/github/fetch/issues/801

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -274,7 +274,15 @@ function Body() {
 
     this.arrayBuffer = function() {
       if (this._bodyArrayBuffer) {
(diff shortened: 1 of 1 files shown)


fetch.js
```diff
@@ -274,7 +274,17 @@ function Body() {
 
     this.arrayBuffer = function() {
       if (this._bodyArrayBuffer) {
-        return consumed(this) || Promise.resolve(this._bodyArrayBuffer)
+        consumed(this)
```
freund17, line 277: Consumed returns either `undefined` or a rejected Promise.

The previous code returned the rejected Promise.
This changes the original behaviour.
Was that intended?

PR #852, opened 2020-10-09
Work-around for core-js issue parsing headers

### LINKED ISSUE #748: IE 11 vs preProcessedHeaders.split(/\r?\n/)
Hello devs, and thank you for this great polyfill.

I am facing a strange bug in IE 11 (on Browserstack, no access to and old Windows system).

In short, the behaviour of [preProcessedHeaders.split(/\r?\n/).forEach(...)](https://github.com/github/fetch/blob/master/fetch.js#L373) seem to be inconsistent between for instance Chrome's last version (my reference point) and IE11 on Browserstack.

**IE 11** -> Incorrect
image

**Chrome** -> Correct
image

Meaning that the polyfill simply doesn't work for me on IE 11.

If you want to try yourself assuming you have an access to an IE 11 browser:
```
'preProcessedHeaders'.split(/\r?\n/)
'preProcessedHeaders\n'.split(/\r?\n/)
'preProcessedHeaders\r'.split(/\r?\n/)
'preProcessedHeaders\r\n'.split(/\r?\n/)
```

Is anyone able to reproduce this?

Thank you for your time in advance.

**Note**
Polyfill version: 3.0.0
Here is the IE11 build I tested with:
image

steveworkman:
Fixes #748 - as discussed in the ticket, this change avoids the use of split via regex, swapping it out for less efficient, but functional ES5 code

//cc @JakeChampion

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -418,14 +418,22 @@ function parseHeaders(rawHeaders) {
   // Replace instances of \r\n and \n followed by at least one space or horizontal tab with a space
   // https://tools.ietf.org/html/rfc7230#section-3.2
   var preProcessedHeaders = rawHeaders.replace(/\r?\n[\t ]+/g, ' ')
-  preProcessedHeaders.split(/\r?\n/).forEach(function(line) {
-    var parts = line.split(':')
-    var key = parts.shift().trim()
-    if (key) {
-      var value = parts.join(':').trim()
-      headers.append(key, value)
-    }
-  })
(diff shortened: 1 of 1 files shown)


eatsjobs:
I can't found this code in the 3.4.1. has been removed?

steveworkman:
This code has not been put into a release yet

JakeChampion:
It's now in 3.5.0 -- https://github.com/github/fetch/releases/tag/v3.5.0

Issue #861, opened 2020-11-16
blob.text is not a function


imtaotao:
Is this a bug ?

```js
fetch(url, conf)
   .then((res) => res.blob())
   .then((blob) => {
      return blob.text().then((code) => {

    })
  }
```

yendi158:
#861

YJ8CJY3MRGJ7MLLT9H3Q:
في الاثنين، ١٦ نوفمبر ٢٠٢٠ ١١:٥٨ ص Arthur <notifications@github.com> كتب:

0531283028:
> [https://github.com/github/fetch/issues/861#issuecomment-728979430]()

HACK3909:
[ in

danielimmke:
This is still an issue

aryan29:
still an issue

wtLau:
still any issue

ZukaiGong:
I triggered this bug today and I don't know why.

JakeChampion:
This project does not provide a polyfill for `blob`

PR #864, opened 2020-11-26
blob.text is not a function #861

meeCnu:
hope this help to chain blob response to support .text to get data as native fetch


meeCnu:
error in code commit

Issue #856, opened 2020-10-24
Request body object is passed as-is


kettanaito:
## Current behavior

When providing a non-stringified object as a request body, `whatwg-fetch` propagates that object to `XMLHttpRequest.send()` as-is.

```js
fetch('/example', { 
  method: 'POST',
  headers: {
    'Content-Type': 'application/json'
  },
  body: { a: 1 }
})
```

## Expected behavior

Expected for this scenario to behave as native `window.fetch`, where such non-stringified request body would be transformed into the `"[object Object]"` string, stating that this is not a valid body per request headers.

## Justification

This raises an incongruence between the polyfill and the implementation it polyfills, producing a different behavior in a browser and NodeJS for the same request issuing code. 

- Originated from https://github.com/mswjs/msw/issues/433

## Reproduction

See this [reproduction repository](https://github.com/kettanaito/whatwg-fetch-json).

kettanaito:
I understand this is a slightly different testing context between the two cases, but I hope it showcases the issue.

## Browser

Unlike `whatwg-fetch`, `window.fetch` in the browser doesn't rely on `XMLHttpRequest` class, so the most objective way to see what it does with the non-stringified body is to inspect the "Request payload' in the network of DevTools.

## NodeJS

In NodeJS I can see that `whatwg-fetch` dispatched `XMLHttpRequest.open` with the request body as-is. By inspecting the polyfill implementation I also conclude that there's no logic that would attempt to stringify the body:

https://github.com/github/fetch/blob/61372fdeb8e6eb59e13da265d518dca21467c1d3/fetch.js#L216-L258

miccycn:
Same issue

JakeChampion:
@kettanaito @miccycn, would either of you be willing to submit a pull-request which fixes this?

kettanaito:
@JakeChampion, sure thing. I'd be glad to ship a fix.

monstro805:
![Copy of repository-open-graph-template](https://user-images.githubusercontent.com/76422372/108236033-4e700b00-70fb-11eb-9ae3-4566526ba92d.png)

kettanaito:
Hey, @JakeChampion. Please, would there be a chance to review my pull request (https://twitter.com/ApiMocking/status/1362084398632488961)? That'd be of huge help.

JakeChampion:
> Hey, @JakeChampion. Please, would there be a chance to review my pull request (#881)? That'd be of huge help.

[...]

PR #881, opened 2020-12-17
Represent non-stringified JSON request body as an [object Object] string

kettanaito:
## Changes

Ensures `request._bodyInit` (the internal representation of a request body) is always set to a `[object Object]` string representation of a body in case it has a non-stringified JSON body and the `Content-Type` header stating an expected JSON body.

## Why

- See #856

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -246,7 +246,9 @@ function Body() {
       this._bodyText = body = Object.prototype.toString.call(body)
     }
 
-    if (!this.headers.get('content-type')) {
+    const contentType = this.headers.get('content-type')
+
+    if (!contentType) {
       if (typeof body === 'string') {
         this.headers.set('content-type', 'text/plain;charset=UTF-8')
       } else if (this._bodyBlob && this._bodyBlob.type) {
@@ -254,6 +256,10 @@ function Body() {
       } else if (support.searchParams && URLSearchParams.prototype.isPrototypeOf(body)) {
         this.headers.set('content-type', 'application/x-www-form-urlencoded;charset=UTF-8')
       }
+    } else if (contentType.includes('json') && typeof this._bodyInit !== 'string') {
+      // Always pass a text representation of a non-stringified JSON body
+      // to `XMLHttpRequest.send` to retain a compatible behavior with the browser.
+      this._bodyInit = this._bodyText
     }
   }
 


kettanaito:
I'd like to ask for help with adding a test for this. Following the existing tests as examples, I've written this test:

```js
test('converts a non-stringified json body to [object Object] string', function() {
  return fetch('/request', {
    method: 'post',
    headers: {
      'content-type': 'application/json'
    },
    body: {stringified: false}
  })
    .then(function(response) {
      return response.json()
    })
    .then(function(request) {
      assert.equal(request.data, '[object Object]')
    })
})
```

However, `request.data` is a public request body that gets correctly set to `[object Object]`. The test doesn't fail when the fix from this pull request is reverted. I can only assume that `request.data` doesn't equal the request body provided to `xhr.send()` at the end of a request:

https://github.com/github/fetch/blob/a8aa427de0ed808ff26c0e3eb2e59c122c44488a/fetch.js#L594

Assertions on `request._bodyInit` fail, that doesn't seem to be public property (dangle implies).

fetch.js
```diff
@@ -246,14 +246,20 @@ function Body() {
...
+
+    if (!contentType) {
       if (typeof body === 'string') {
         this.headers.set('content-type', 'text/plain;charset=UTF-8')
       } else if (this._bodyBlob && this._bodyBlob.type) {
         this.headers.set('content-type', this._bodyBlob.type)
       } else if (support.searchParams && URLSearchParams.prototype.isPrototypeOf(body)) {
         this.headers.set('content-type', 'application/x-www-form-urlencoded;charset=UTF-8')
       }
+    } else if (contentType.includes('json') && typeof this._bodyInit !== 'string') {
```
kettanaito, line 259: This isn't a reliable JSON request body content-type header check. Would you recommend to provide `application/json; charset=UTF-8` instead? 

I'm thinking that this behavior should also affect _any_ JSON-like content-type header (i.e. `application/hal+json`).

lektor2034:
thk sir

fetch.js
```diff
@@ -246,14 +246,20 @@ function Body() {
       this._bodyText = body = Object.prototype.toString.call(body)
```
kettanaito, line 246: Alternatively, we can set `this._bodyInit` to equal `this._bodyText` in this closure. I'm not sure about the implications of doing this, so would appreciate a feedback.

whiteslack: whiteslack

whiteslack (approved):
whiteslack

Toppole69 (approved):

Toppole69 (approved):

SonyViao (approved):

SonyViao (approved):

SonyViao (approved):

fetch.js
```diff
@@ -246,14 +246,20 @@ function Body() {
...
     }
 
-    if (!this.headers.get('content-type')) {
```
marcin543, line 249: - [ ]

fetch.js
```diff
@@ -246,14 +246,20 @@ function Body() {
...
+    const contentType = this.headers.get('content-type')
```
marcin543, line 249: - [ ]

fetch.js
marcin543: - [ ]

fetch.js
```diff
@@ -246,14 +246,20 @@ function Body() {
...
       } else if (this._bodyBlob && this._bodyBlob.type) {
         this.headers.set('content-type', this._bodyBlob.type)
       } else if (support.searchParams && URLSearchParams.prototype.isPrototypeOf(body)) {
         this.headers.set('content-type', 'application/x-www-form-urlencoded;charset=UTF-8')
       }
+    } else if (contentType.includes('json') && typeof this._bodyInit !== 'string') {
+      // Always pass a text representation of a non-stringified JSON body
+      // to `XMLHttpRequest.send` to retain a compatible behavior with the browser.
+      this._bodyInit = this._bodyText
     }
```
marcin543, line 263: - [ ]

marcin543 (approved):

skkienn27angela (approved):
>

skkienn27angela:
``

JakeChampion (approved):

JakeChampion:
This is released in v3.6.0 -- https://github.com/github/fetch/releases/tag/v3.6.0

NeverLiez (commented):
B

Apple1D (commented):
- [ ]

PR #887, opened 2021-01-12
add the URL to the failed request error messages

oscarmarimrf:
It would be useful to have the requested URL in error messages.
This PR adds it to the failed request errors.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -526,13 +526,13 @@ export function fetch(input, init) {
 
     xhr.onerror = function() {
       setTimeout(function() {
-        reject(new TypeError('Network request failed'))
+        reject(new TypeError('Network request failed: ' + request.url))
       }, 0)
     }
 
     xhr.ontimeout = function() {
       setTimeout(function() {
-        reject(new TypeError('Network request failed'))
+        reject(new TypeError('Network request failed: ' + request.url))
       }, 0)
     }
 


fetch.js
```diff
@@ -526,13 +526,13 @@ export function fetch(input, init) {
...
       setTimeout(function() {
-        reject(new TypeError('Network request failed'))
+        reject(new TypeError('Network request failed: ' + request.url))
       }, 0)
     }
 
     xhr.ontimeout = function() {
       setTimeout(function() {
-        reject(new TypeError('Network request failed'))
+        reject(new TypeError('Network request failed: ' + request.url))
```
kettanaito, line 535: While this is a noble change to improve the debugging quality, such error would be incompatible with a TypeError thrown by `window.fetch`:

```js
fetch('http://foo.bar')
// Uncaught (in promise) TypeError: Failed to fetch
//    at <anonymous>:1:1
```

This repository is a polyfill for `window.fetch`, which means that it treats `window.fetch` as the source of truth and a specification to look up to.

kettanaito (changes requested):
Thank you for this addition, but it would break compatibility with `window.fetch` in terms of a `TypError` message this polyfill would produce.

fetch.js
oscarmarimrf: That makes sense, thank you very much for the detailed answer.

PR #914, opened 2021-02-18
Fix MSIE compatibility

alexbakum:
Version 3.6.0 (https://github.com/github/fetch/pull/881/files#diff-3fc6f1265a180d15035618e36ca1da88d092bacb4fc10a423938fa62e4f5f58fR259) broke MSIE compatibility.
`String.prototype.includes` is not available within any version of MSIE – [see compatibility table on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#browser_compatibility)

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -246,7 +246,7 @@ function Body() {
       this._bodyText = body = Object.prototype.toString.call(body)
     }
 
-    const contentType = this.headers.get('content-type')
+    var contentType = this.headers.get('content-type')
 
     if (!contentType) {
       if (typeof body === 'string') {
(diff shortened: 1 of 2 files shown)


alexbakum:
@JakeChampion please take a look

dasa:
Should this be set back too?

https://github.com/github/fetch/blob/90fb680c1f50181782f276122c1b1115535b1603/.eslintrc.json#L3

JakeChampion:
> Should this be set back too?
> 
> https://github.com/github/fetch/blob/90fb680c1f50181782f276122c1b1115535b1603/.eslintrc.json#L3

it can't because the file use `export` syntax

dasa:
Would it work if `sourceType` was set to "module"? https://eslint.org/docs/user-guide/configuring/language-options#specifying-parser-options

PR #916, opened 2021-02-18
use var instead of const

JakeChampion:


JakeChampion:
closing in favour of https://github.com/github/fetch/pull/914

PR #944, opened 2021-03-24
Delete karma-worker.config.js

Tammie28:


JakeChampion:
We won't be deleting this file as it's used in `npm test`.

PR #956, opened 2021-04-19
Get the proper global object depending on environment to check for existing fetch

jcs224:
This might be more of a conversation than a pull request, but I want to attempt to address an issue that's been noticed in a dependency, `cross-fetch`, that this might help solve.

As it stands, `XMLHttpRequest` doesn't exist in some runtimes, such as Deno or Cloudflare Workers. Tweaking this library might have a positive cascade of solving some issues, such as these:

https://github.com/supabase/supabase-js/issues/154
https://github.com/lquixada/cross-fetch/issues/78
https://github.com/supabase/supabase-js/issues/161

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -605,9 +605,19 @@ export function fetch(input, init) {
 
 fetch.polyfill = true
 
-if (!global.fetch) {
-  global.fetch = fetch
-  global.Headers = Headers
-  global.Request = Request
-  global.Response = Response
+var getGlobal = function () {
+  if (typeof globalThis !== 'undefined') { return globalThis }
+  if (typeof self !== 'undefined') { return self }
+  if (typeof window !== 'undefined') { return window }
+  if (typeof global !== 'undefined') { return global }
+  throw new Error('unable to locate global object')
+}
+
+var globals = getGlobal()
+
+if (!globals.fetch) {
+  globals.fetch = fetch
+  globals.Headers = Headers
+  globals.Request = Request
+  globals.Response = Response
 }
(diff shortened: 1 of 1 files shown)


JakeChampion:
@jcs224, we already have similar code in-place which only adds the polyfills to the global object if the native `fetch` does not exist:
https://github.com/github/fetch/blob/master/fetch.js#L600-L605

This package wasn't written with those runtimes (deno, cloudflare workers) in mind, is there a way to run our test suite in those runtimes?

jcs224:
Thanks for the feedback. Now that you mention that snippet, I think the problem is that the `global` object doesn't exist in environments outside of Node, which is the reason it's breaking outside of Node and normal browser environments. `globalThis` is supposed to be the way around it, but it's not available in some older environments. I made an update that accounts for the variations in the global namespace, taken directly from MDN:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#search_for_the_global_across_environments

I'm not sure how to test this in the other environments, to be honest. I only just started getting into Deno and just heard about Cloudflare Workers, so more guidance might be needed there. At the very least, don't what to disrupt the library for anyone else using it.

jcs224:
@JakeChampion I changed the title of the PR to more accurately reflect the actual change I'm proposing.

JakeChampion:
@jcs224 we went with a different approach then the one you proposed in this pull-request

We now set the variable to an empty object if none of the expected globalThis names work -- https://github.com/github/fetch/pull/967

jcs224:
@JakeChampion excellent, thanks!

PR #967, opened 2021-05-08
Compatible global equals to the false

cn-yezi:
In some special environments, such as webview in certain scenes.
`globalThis` `self` `global` Neither holds, var global will be equal to false.
The following code will throw an error.

```javascript
searchParams: 'URLSearchParams' in global,
iterable: 'Symbol' in global && 'iterator' in Symbol,

// 'URLSearchParams' in false
```

**TypeError: Cannot use 'in' operator to search for 'URLSearchParams' in false**

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,7 +1,8 @@
 var global =
   (typeof globalThis !== 'undefined' && globalThis) ||
   (typeof self !== 'undefined' && self) ||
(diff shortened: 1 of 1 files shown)


JakeChampion (approved):

cn-coderl:
@JakeChampion Excuse me, the workflow needs approval.

PR #1016, opened 2021-08-25
Update bower.json

frook58:


Shah8018 (commented):
New era for cryptoC

Shah8018 (commented):
New era for cryptoC

PR #1023, opened 2021-08-29
Create node2.js.yml

angie1148:
- name: Set up gcloud Cloud SDK environment
  # You may pin to the exact commit or the version.
  # uses: google-github-actions/setup-gcloud@94337306dda8180d967a56932ceb4ddcf01edae7
  uses: google-github-actions/setup-gcloud@v0.2.0
  with:
    # Version of the gcloud SDK to install. If unspecified or set to "latest",
the latest available gcloud SDK version for the target platform will be
installed. Example: "290.0.1".
    version: # optional, default is latest
    # Service account email address to use for authentication. This is required
for legacy .p12 keys but can be omitted for .json keys. This is usually of
the format <name>@<project-id>.iam.gserviceaccount.com.
    service_account_email: # optional
    # Service account key to use for authentication. This should be the JSON
formatted private key which can be exported from the Cloud Console. The
value can be raw or base64-encoded.
    service_account_key: # optional
    # ID of the Google Cloud project. If provided, this will configure gcloud to
use this project ID by default for commands. Individual commands can still
override the project using the --project flag which takes precedence.
    project_id: # optional
    # Export the provided credentials as Google Default Application Credentials.
This will make the credentials available to later steps via the
GOOGLE_APPLICATION_CREDENTIALS environment variable. Future steps that
consume Default Application Credentials will automatically detect and use
these credentials.
    export_default_credentials: # optional

### DIFF
--- a/.github/workflows/node2.js.yml
+++ b/.github/workflows/node2.js.yml
@@ -1,0 +1,31 @@
+# This workflow will do a clean install of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


angie1148 (commented):
Looks good

angie1148 (commented):
Great

PR #1081, opened 2021-11-03
Remove error that causes ie11 to be unresponsive

robertlandreaux:
The headers of requests to Zendesk and Sentry contain colons, e.g.

:authority
:scheme
:method
:path

Those headers cause this error to be thrown. This error is thrown continuously because Sentry's JavaScript continues to send the error to Sentry. Every fetch request to Sentry causes another error. When the requests are in an infinite loop, IE 11 becomes unresponsive.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -50,9 +50,6 @@ function normalizeName(name) {
   if (typeof name !== 'string') {
     name = String(name)
   }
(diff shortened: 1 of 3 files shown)


robertlandreaux:
Sorry. I meant to make this branch merge into our fork.

JakeChampion:
@robertlandreaux That wouldn't work with the native fetch implementation, it will throw a TypeError within the Headers constructor

Issue #1371, opened 2023-08-02
File fetching broken since commit 0c1d2b9


ldaudet:
Hi there!

Thanks a lot for the work on this library! I am using it in a react native app to run code like this: https://snack.expo.dev/@bacon/documentpicker-to-blob (line 24), to grab an image file that I will then upload to the backend.

This worked for the longest time, but after uploading to the latest version of expo, it also uploaded the version of react native, which in turned updated our version of fetch. The problem is that with the latest version of fetch, the fetch of file doesn't work anymore, as described here: https://github.com/facebook/react-native/issues/38625#issuecomment-1651430009

After looking at the code, I realized that the issue was actually do to the work on commit 0c1d2b9 where you added a check for the return code to be between certain values:
```
  if (this.status < 200 || this.status > 599) {
    throw new RangeError("Failed to construct 'Response': The status provided (0) is outside the range [200, 599].")
  }
```
When removing this check, everything works just fine with fetching the file, even with the lastest version of fetch, as the fetch of the a file return a status 0.

Would it be possible to either remove the check, or update the code to get a 200 status on successfully fetching a file?

Thanks a lot

mercpls:
@ldaudet Still waiting on a response but there is a PR based on this suggestion now.

meteorlxy:
According to [fetch spec - Schema fetch](https://fetch.spec.whatwg.org/#scheme-fetch):

> "file"
> For now, unfortunate as it is, file: [URLs](https://www.w3.org/TR/pub-manifest/#dfn-url) are left as an exercise for the reader.
>
> When in doubt, return a [network error](https://fetch.spec.whatwg.org/#concept-network-error).

I'm not sure what's the exact meaning of the spec - Should fetch supports file uri or not?

Any native speaker or spec expert can answer it? 🤔

JakeChampion:
Fixed via #1375 👍

mercpls:
> According to [fetch spec - Schema fetch](https://fetch.spec.whatwg.org/#scheme-fetch):
> 
> > "file"
> > For now, unfortunate as it is, file: [URLs](https://www.w3.org/TR/pub-manifest/#dfn-url) are left as an exercise for the reader.
> > When in doubt, return a [network error](https://fetch.spec.whatwg.org/#concept-network-error).
> 
> I'm not sure what's the exact meaning of the spec - Should fetch support file uri or not?
> 
> Any native speaker or spec expert could help answer it? 🤔

[...]

PR #1375, opened 2023-08-10
Fix - File fetching broken since commit 0c1d2b9

mercpls:
#1371 
As noted in 1371, and other discussion outside of whatwg/fetch, it's been found that adding the RangeError for statuses outside of 200-599 causes local File fetches to fail.

i.e.
fetch("file://some/local/path")

This PR adds a simple check for the file path before returning the response, it could alternatively be added to the 200/500 check as a not condition.

Not sure if this has the potential to break other things, or if there's a better way to handle the RangeError issue with `file://` not setting a status.

Test: In it's current state, I'm unsure how to test this, as most browsers block the use of `file://` and any systems that allow may return a status, and may not.

Thanks

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -542,6 +542,12 @@ export function fetch(input, init) {
         statusText: xhr.statusText,
         headers: parseHeaders(xhr.getAllResponseHeaders() || '')
       }
+      // This check if specifically for when a user fetches a file locally from the file system
+      if (request.url.startsWith('file://')) {
+        options.status = 200;
+      } else {
+        options.status = xhr.status;
+      }
       options.url = 'responseURL' in xhr ? xhr.responseURL : options.headers.get('X-Request-URL')
       var body = 'response' in xhr ? xhr.response : xhr.responseText
       setTimeout(function() {
--- a/fetch.js
+++ b/fetch.js
@@ -470,7 +470,9 @@ export function Response(bodyInit, options) {
 
   this.type = 'default'
   this.status = options.status === undefined ? 200 : options.status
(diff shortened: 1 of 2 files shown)


bogdanbpeterson:
Confirm, RangeError is freaking me out after I've upgraded to expo@49
Error output:
```
ERROR  RangeError: Failed to construct 'Response': The status provided (0) is outside the range [200, 599]., js engine: hermes
```
Code that causes this error:
```typescript
await (await fetch(fileUrl)).blob())
```

mercpls:
> Confirm, RangeError is freaking me out after I've upgraded to expo@49 Error output:
> 
> ```
> ERROR  RangeError: Failed to construct 'Response': The status provided (0) is outside the range [200, 599]., js engine: hermes
> ```
> 
> Code that causes this error:
> 
> ```ts
> await (await fetch(fileUrl)).blob())
> ```

___

@bogdanbpeterson A temporary fix is to create your own XMLHttpRequest, just like this library does, except you would handle the errors on your own so that this issue doesn't happen.

You can also rollback to the commit of the whatwg/fetch version before the RangeError was introduced.
```ts
const uriToBlob = (uri: string): Promise<Blob> => {
  return new Promise((resolve, reject) => {
    const xhr = new XMLHttpRequest();
    xhr.onload = () => resolve(xhr.response);
    xhr.onerror = () => reject(new Error('uriToBlob failed'));
    xhr.responseType = 'blob';
    xhr.open('GET', uri, true);
    xhr.send(null);
  });
};
```

JakeChampion (approved):

PR #1387, opened 2023-09-25
Hash pin actions used on workflows and enable dependabot

### LINKED ISSUE #1386: Hash pin actions used on workflows and enable depenadabot
Hi again, I'd like to suggest another security practice that helps increasing CI stability and prevent tag renaming attack.

This practice is recommended by both the [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies) and the [github security hardening][github-hashpin].

It is really important to ensure privileged actions and/or third-party actions are always hash-pinned. 
  
Considering dependabot grouping PRs, it is now much more easier to ensure keeping actions immutable, but still up to date.

Any concerns or questions about it, let me know! 
Thanks.

### Additional Context

A [tag renaming attack][tag-renaming] is a type of attack whereby an attacker:
- Hijack an action.
- Upload a malicious version.
- Replace existing tags with malicious versions.

[tag-renaming]: https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash
[github-hashpin]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

joycebrum:
Closes #1386 

Hash pinned actions used on workflows and enabled the dependabot to group PRs and max of 1 for week.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
       github-actions:
         patterns:
           - "*"
-    # open-pull-requests-limit: 3
+    open-pull-requests-limit: 3
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
       github-actions:
         patterns:
           - "*"
-    open-pull-requests-limit: 3
+    open-pull-requests-limit: 1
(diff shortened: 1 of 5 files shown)


JakeChampion (approved):
Thank you

PR #1404, opened 2023-11-10
Replaced startsWith with indexOf to Avoid Additional Polyfill Dependencies

elasim:
PR will resolve https://github.com/JakeChampion/fetch/issues/1403

Since startsWith has a similar level of support as fetch, we have replaced it with indexOf to ensure that the module does not require additional polyfills.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -543,7 +543,7 @@ export function fetch(input, init) {
       }
       // This check if specifically for when a user fetches a file locally from the file system
       // Only if the status is out of a normal range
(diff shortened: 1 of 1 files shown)


JakeChampion (approved):

bergmark:
@hishamhalil Could you provide context on why this was reverted please?

bergmark:
Could we get a release for this please?

(Deleted my first comment, I thought it had been reverted in this repository...)

elasim:
Looks like I didn't realize that using conventional commit triggers automatic release.

@JakeChampion
How should I go about deploying this?
Is it better for us to wait for other contributions?

PR #1472, opened 2024-10-15
Implement Set-Cookie handling

### LINKED ISSUE #1384: `Set-Cookie` headers should not be combined / Support `Headers.getSetCookie()` method
The Fetch Standard defines [special behaviour for `Set-Cookie` headers](https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine):

> To **sort and combine** a [header list](https://fetch.spec.whatwg.org/#concept-header-list) list, run these steps. They return a [header list](https://fetch.spec.whatwg.org/#concept-header-list).
> 
> 1. Let headers be a [header list](https://fetch.spec.whatwg.org/#concept-header-list).
> 2. Let names be the result of [convert header names to a sorted-lowercase set](https://fetch.spec.whatwg.org/#convert-header-names-to-a-sorted-lowercase-set) with all the [names](https://fetch.spec.whatwg.org/#concept-header-name) of the [headers](https://fetch.spec.whatwg.org/#concept-header) in list.
> 3. [For each](https://infra.spec.whatwg.org/#list-iterate) name of names:
>     1. If name is \``set-cookie`\`, then:
>         1. Let values be a list of all [values](https://fetch.spec.whatwg.org/#concept-header-value) of [headers](https://fetch.spec.whatwg.org/#concept-header) in list whose [name](https://fetch.spec.whatwg.org/#concept-header-name) is a [byte-case-insensitive](https://infra.spec.whatwg.org/#byte-case-insensitive) match for name, in order.
>         2. [For each](https://infra.spec.whatwg.org/#list-iterate) value of values:
>             1. [Append](https://infra.spec.whatwg.org/#list-append) (name, value) to headers.
>     2. Otherwise:
>         1. Let value be the result of [getting](https://fetch.spec.whatwg.org/#concept-header-list-get) name from list.
>         2. Assert: value is non-null.
>         3. [Append](https://infra.spec.whatwg.org/#list-append) (name, value) to headers.
> 4. Return headers.

This is because the values of `set-cookie` headers can contain commas, such as in the case of a cookie with an expiry time (e.g. `cookie-name=cookie-value; Path=/; Expires=Thu, 21 Sep 2023 06:12:39 GMT; Secure`). Combining them with commas can change their semantics.

This library currently seems to ignore this part of the spec, and combine `set-cookie` values. For example, this is the current behaviour of `whatwg-fetch`:

```js
const { Headers } = require("whatwg-fetch");

const headers = new Headers();
headers.append("Set-Cookie", "x=x");
headers.append("Set-Cookie", "y=y");

console.log(JSON.stringify([...headers.entries()])); // outputs [["set-cookie","x=x, y=y"]]
```

In comparison, Chrome and Firefox follow the spec (Safari doesn't):

Chrome output

Firefox output

This seems to be a [pretty recent addition](https://github.com/whatwg/fetch/pull/1346) to the spec (2001ish). There's also a new method [Headers: getSetCookie() method](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) added around the same time which is missing from `whatwg-fetch`:

Chrome getSetCookie output

For context, I'm using `whatwg-fetch` as a polyfill for running Jest tests that use Mock Service Worker (MSW), similar to [this example](https://github.com/mswjs/examples/tree/master/examples/with-jest) over in the MSW project. I noticed this issue because I was setting multiple cookies with different options, and the combining of the cookie values was causing strange behaviour in my tests.

LinusU:
Fixes #1384 

Specification: https://fetch.spec.whatwg.org/#dom-headers-getsetcookie

Why is this needed since the browsers doesn't expose `Set-Cookie`? React Native is using this polyfill to provide `fetch`, and it does expose the `Set-Cookie` header.

-----

An alternative implementation would be to store the `set-cookie` headers separately from the `map` variable. Please let me know if you would prefer that, and I'll update the code.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -107,8 +107,11 @@ export function Headers(headers) {
 Headers.prototype.append = function(name, value) {
   name = normalizeName(name)
   value = normalizeValue(value)
-  var oldValue = this.map[name]
-  this.map[name] = oldValue ? oldValue + ', ' + value : value
+  if (this.has(name)) {
+    this.map[name].push(value)
+  } else {
+    this.map[name] = [value]
+  }
 }
 
 Headers.prototype['delete'] = function(name) {
@@ -117,7 +120,7 @@ Headers.prototype['delete'] = function(name) {
 
 Headers.prototype.get = function(name) {
   name = normalizeName(name)
-  return this.has(name) ? this.map[name] : null
+  return this.has(name) ? this.map[name].join(', ') : null
 }
 
 Headers.prototype.has = function(name) {
@@ -125,13 +128,19 @@ Headers.prototype.has = function(name) {
 }
 
 Headers.prototype.set = function(name, value) {
-  this.map[normalizeName(name)] = normalizeValue(value)
+  this.map[normalizeName(name)] = [normalizeValue(value)]
 }
 
 Headers.prototype.forEach = function(callback, thisArg) {
   for (var name in this.map) {
     if (this.map.hasOwnProperty(name)) {
-      callback.call(thisArg, this.map[name], name, this)
+      if (name === 'set-cookie') {
+        for (var index in this.map[name]) {
+          callback.call(thisArg, this.map[name][index], name, this)
+        }
+      } else {
+        callback.call(thisArg, this.map[name].join(', '), name, this)
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


opmdod360-max (approved):
Is this better

opmdod360-max:
> Fixes #1384
> 
> Specification: https://fetch.spec.whatwg.org/#dom-headers-getsetcookie
> 
> Why is this needed since the browsers doesn't expose `Set-Cookie`? React Native is using this polyfill to provide `fetch`, and it does expose the `Set-Cookie` header.
> 
> An alternative implementation would be to store the `set-cookie` headers separately from the `map` variable. Please let me know if you would prefer that, and I'll update the code.

There you go

PR #1475, opened 2024-12-18
For request failures and timeouts, include the URLs in the errors

giles-v:
When working with complex Jest tests which mount large component trees, it can be difficult to nail down which request is causing test failures with the current message. Echo the specific method and URL into the Error, so that debugging is easier.

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -558,13 +558,13 @@ export function fetch(input, init) {
 
     xhr.onerror = function() {
       setTimeout(function() {
-        reject(new TypeError('Network request failed'))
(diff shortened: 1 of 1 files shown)


fetch.js
```diff
@@ -558,13 +558,13 @@ export function fetch(input, init) {
 
     xhr.onerror = function() {
       setTimeout(function() {
-        reject(new TypeError('Network request failed'))
+        reject(new TypeError('Network request ('+request.method+' '+request.url+') failed'))
```
giles-v, line 561: These would be neater as string template literals, but since those were not currently in use in the codebase, I opted to match the existing style of plain strings.

